PR #65, opened 2013-10-23
Avoid simplexml_load_file warning when config path is a directory

paxal:
When using a directory for phpunit configuration path, with WARNING display enabled, it creates a multiple warnings on console output, thus messing report up.

I just added a new test to check if file is a directory, and if so then do not call the simple_load_file function on the directory.

### DIFF
--- a/src/ParaTest/Runners/PHPUnit/Options.php
+++ b/src/ParaTest/Runners/PHPUnit/Options.php
@@ -170,14 +170,30 @@ class Options
     protected function getConfigurationPath($filtered)
     {
         if(isset($filtered['configuration']))
+            return $this->getDefaultConfigurationForPath($filtered['configuration'], $filtered['configuration']);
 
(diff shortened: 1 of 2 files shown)


brianium:
@julianseeger @paxal  do you have any more information on this? is this still an issue? if the PR is dead in the water can the PR be closed and a related issue be opened?

julianseeger:
Maybe you could open an issue and label it as an ```unconfirmed``` ```bug```.
I don't remember any fix in this are so it's probably still there.

If @paxal wants to finish this: your tests were failing because of an instability in our performance-tests (that has been fixed a while ago). So if you would rebase / merge your fork with the current paratest master it should work.
In terms of the behavior, the preferred behavior is to imitate the behavior of phpunit. I think it is sufficient if you have a test for each of the cases you mentioned: nothing specified + phpunit.xml, nothing specified + phpunit.xml.dist, directory specified + phpunit.xml, directory specified + phpunit.xml.dist.

PR #74, opened 2014-01-30
Test loading improvements

oker1:
Improved test loading to be more robust than the regexps used to parse the source files. We needed this because we have tests with different filenames than the class in them with multiline class statements (see the testcase for this). We could improve the regexp to handle these cases, but I think this solution is more robust. Also it can be extended easily to support test files containing multiple test classes.

I'm trying to make paratest work on our test suite containing roughly ten thousand tests so I expect more phpunit compatibility issues to fix :)


Slamdunk:
This will not be supported anymore in the near future, see https://github.com/paratestphp/paratest/pull/422

PR #96, opened 2014-05-15
[Bugfix: windows] Allow paratest to work even when paths contain spaces

cs462-unikent:
Hello,

When attempting to run paratest on windows I was repeatedly getting `"Log file XYZ is empty. This means a PHPUnit process has crashed.`. After a bit of debugging it turns out this was due to phpunit being passed options (such as paths) that contained spaces.

I've added a fix that effectively just encases the option values in quotes in order to resolve this issue. Everything appears to be working normally for me now so i thought I may I may as well put in a pull with the fix.

Thanks,
Carl

### DIFF
--- a/src/ParaTest/Runners/PHPUnit/ExecutableTest.php
+++ b/src/ParaTest/Runners/PHPUnit/ExecutableTest.php
@@ -179,8 +179,8 @@ abstract class ExecutableTest
     {
         // TODO: this should use a CommandBuilder
         $command = $binary;
-        foreach($options as $key => $value) $command .= " --$key %s";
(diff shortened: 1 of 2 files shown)


brianium:
@cs462-unikent any word on this? still having issues - @julianseeger @nubs @aik099 anything else we can do to move this along? Is it time to close this?

julianseeger:
It is probably fixed because the symfony process builder has a solid quoting.
If nobody with a windows-rig is willing to confirm that this issue still exists, I assume that it is fixed.

brianium:
Not hearing anything on this, so I too will assume its fixed. If we have a windows user with this issue again, I am sure we will hear about it :) - I know @chbiel is working on windows, and things seem to be going well for him (aside from WrapperRunner support of course).

PR #127, opened 2014-12-23
Fix handling of '&' symbols in JUnit XML

RojerUA:
Without this change whenever  URLs with '&' symbol appears in phpunit output, it triggers warning and URL is corrupted in Jenkins.

Example URL:
http://test.site.com?key=1419318842&ip=0i1BxHBaShQueFNYFQaA3IbyAJC3HncWngeY2hZdEy3bSTFOkeXUzbusMUyiisAPAgoc50Ms-wuTS_rWwUHLvPNzFchl8ebXm3ljF83mfzxX98SaPLzWtoyMQauIwHWqKknjtRxRI1flde3ovN7kdQ&im=image/png&o=&cd=inline

Error in console:
 [exec] #2 /var/www/optimalprint-trunk/internal/SL/Test/Functional/Page/Category/CategoryDesktopPHP Warning:  DOMDocument::createElement(): unterminated entity reference ip=0i1BxHBaShQueFNYFQaA3IbyAJC3HncWngeY2hZdEy3bSTFOkeXUzbusMUyiisAPAgoc50Ms-wuTS_rWwUHLvPNzFchl8ebXm3ljF83mfzxX98SaPLzWtoyMQauIwHWqKknjtRxRI1flde3ovN7kdQ&im=image/png&o=&cd=inline

### DIFF
--- a/src/ParaTest/Logging/JUnit/Writer.php
+++ b/src/ParaTest/Logging/JUnit/Writer.php
@@ -147,7 +147,7 @@ class Writer
     protected function appendDefects($caseNode, $defects, $type)
     {
         foreach($defects as $defect) {
-            $defectNode = $this->document->createElement($type, $defect['text'] . "\n");
+            $defectNode = $this->document->createElement($type, htmlentities($defect['text']) . "\n");
             $defectNode->setAttribute('type', $defect['type']);
             $caseNode->appendChild($defectNode);
         }


src/ParaTest/Logging/JUnit/Writer.php
```diff
@@ -147,7 +147,7 @@ protected function appendCase($suiteNode, TestCase $case)
     protected function appendDefects($caseNode, $defects, $type)
     {
         foreach($defects as $defect) {
-            $defectNode = $this->document->createElement($type, $defect['text'] . "\n");
+            $defectNode = $this->document->createElement($type, htmlentities($defect['text']) . "\n");
```
jeffchan: While this works for the case when there's only `&` in the string, this breaks other cases when there are other HTML entities.

For example: if the text has `—`, then this line change converts it to `&mdash;` which the breaks the junit.xml parsing.

jeffchan: A better fix is to use `htmlspecialchars` which only escapes the necessary characters.

jeffchan: See #296

PR #135, opened 2015-01-02
Error message when empty

Cameronjmayfield:
The reason the previous pull request (https://github.com/brianium/paratest/pull/133) was failing is because when the tests are run with multiple, comma separated groups, the framework would pull all tests to execute. This leaves an empty log file, which is the symptom I was using to diagnose the issue.

The newest change updates the command to get test methods to iterate over comma separated values passed into the annotations.

### DIFF
--- a/src/ParaTest/Logging/JUnit/Reader.php
+++ b/src/ParaTest/Logging/JUnit/Reader.php
@@ -72,6 +72,17 @@ class Reader extends MetaProvider
     }
 
     /**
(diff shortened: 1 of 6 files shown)


julianseeger:
Ok, this is actually **"Error message when [suite is] empty" + "No longer runs unselected groups when multiple are selected"** (just to point that out).
Looks good to me, to travis, to scrutinizer and is well tested => merged.

Thanks!

PR #140, opened 2015-01-14
ignore testfiles without classes

julianseeger:
ignore testfiles without classes
+ fix for filenamed based classloading order when using namespaces

this fixes https://github.com/brianium/paratest/issues/139 and a bug in https://github.com/brianium/paratest/issues/134 (it didn't work with namespaces)

FTR: these two fixes make the whole monolog testsuite to run with paratest


brianium:
@julianseeger nice! shall I merge?

julianseeger:
please ;)

padraic:
Thanks guys!

Issue #142, opened 2015-01-21
Support for --testsuite phpunit parameter


chbiel:
It would be nice to specify the --testsuite parameter and execute the tests in the testsuite in parallel.

E.g.
```xml
<testsuite name="MySuite">
    <file>MyFirstTest.php</file>
    <file>MySecondTest.php</file>
</testsuite>
```

I also have other suites so I only what to execute that one:
paratest --testsuite=MySuite --processes=2

Would be very nice :)

chbiel:
added pull request to add this feature:
https://github.com/brianium/paratest/pull/144

kick-the-bucket:
PHPUnit actually support filtering by multiple testsuites separated by comma which is currently not supported by paratest.
Maybe should this be reopened or a new issue created?

PR #144, opened 2015-01-29
add support for testsuite parameter (#142)

chbiel:
make it possible to spedify a testsuite within phpunit.xml with <file> attributes and execute the specified <file>'s in parallel

### DIFF
--- a/src/ParaTest/Console/Testers/PHPUnit.php
+++ b/src/ParaTest/Console/Testers/PHPUnit.php
@@ -43,6 +43,7 @@ class PHPUnit extends Tester
             ->addOption('stop-on-failure', null, InputOption::VALUE_NONE, 'Don\'t start any more processes after a failure.')
             ->addOption('log-junit', null, InputOption::VALUE_REQUIRED, 'Log test execution in JUnit XML format to file.')
             ->addOption('colors', null, InputOption::VALUE_NONE, 'Displays a colored bar as a test result.')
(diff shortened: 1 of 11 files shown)


julianseeger:
Nice, thanks for this.

I'm just wondering about files but not directories. Your patch supports suites with files but doesn't support directories, does it? (https://phpunit.de/manual/current/en/organizing-tests.html#organizing-tests.xml-configuration)
Actually, I haven't seen a testsuite with files in practice. Is it much harder to add directory support? If not, it would be nice if you could add it.

chbiel:
No problem, I need this for a project with many legacy tests with messed up test structure ;)

I added directory support and (apparently) support for mixed file/directory suites.
I wasn't sure if I should make it complicated with another option to filter duplicates (when mixing files and directories) so I forced, that no duplicate tests are allowed when using --testsuite parameter

julianseeger:
>I wasn't sure if I should make it complicated with another option to filter duplicates (when mixing files and directories) so I forced, that no duplicate tests are allowed when using --testsuite parameter.

Sounds good. Getting an error is better than having duplicates because of a senseless config and not noticing it.

>It's important to mention, that the <file>'s are not executed in order as mentioned in the phpunit docs. For paratests it is only to add specific files to a testsuite that will be also executed.

The order is always mixed up when parallelizing the tests, so +1 for that.

Thanks again and this in going to be version 0.10.0

PR #147, opened 2015-02-04
Allow custom runners

chbiel:
This pull requests allows the developer to use a custom runner that extends BaseRunner

For me this is a semi-workaround for the windows wrapperrunner problem, because I can provide my own runner with pre-post scripts per token. but i will have a closer look into the windows issues when there is time

also change the tokens from numeric to uniqueid's because that is more generic and you can e.g. do something like creating a prepared database for every token before running the actual test (this custom runner is not included)

### DIFF
--- a/src/ParaTest/Console/Testers/PHPUnit.php
+++ b/src/ParaTest/Console/Testers/PHPUnit.php
@@ -66,6 +66,17 @@ class PHPUnit extends Tester
         if ($input->getOption('runner') === 'WrapperRunner') {
             $runner = new WrapperRunner($this->getRunnerOptions($input));
         } else {
+            if ($input->getOption('runner') !== '') {
+                // because we want to have to bootstrap script inherited before check/initialization
+                $runnerOption = $this->getRunnerOptions($input);
+                $runnerClass = $input->getOption('runner');
+                if (class_exists($runnerClass)) {
+                    $runner = new $runnerClass($runnerOption);
+                }
+            }
+        }
+
+        if (!isset($runner)) {
             $runner = new Runner($this->getRunnerOptions($input));
         }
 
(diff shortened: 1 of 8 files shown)


julianseeger:
FTR: The HHVM builds had an error but a rerun of the travis build succeeded.

Could you please provide more details what the uniqueid tokens are good for?
I'm using the tokens in multiple projects to prepare databases with names based on the tokens and never noticed any issues.
Changing from small numbers to 13 alphanumeric chars MIGHT eventually break some setups. Is it really an issue or just because you like uniqueid? ;)

chbiel:
sure.
I got to the following problem with normal incrementing integers (in an extended CI process):

expecting you have one big mysql database that holds all databases for your unit test for all deployment stages (to not mess up the actual application database).
when you run unittests for different stages of the ci process with paratest in parallel, every stage want to create the database as (e.g.)
db_tests_1
db_tests_2
db_tests_3
etc.
so executing 2 paratest's in parallel will fail because they want use the same databases. uniqueid ensure that this does not happen :)

julianseeger:
Interesting scenario. Oh and feel free to share some runners with us if you think they add value and are not too specific to your environment.

Would you mind adding a test for the PHPUnit-Tester? Maybe you could write a MockRunner and check it was called or something like that, to make sure nobody brakes this feature in the future.

chbiel:
Yes I can do that later today :)

chbiel:
I added a test that checks if the run method of the running class is invoked.
This could not be done through $this->once() because the actual thing to check is in the subprocess.

So, I added the possibility to catch the sub-processes output via callback

julianseeger:
Thanks for the contribution.
This is release 0.11.0

PR #148, opened 2015-02-05
Nested testsuites support

chbiel:
For better structure in phpunit.xml I added the possibility to use nested <testsuite> elements.

for example this is helpful in such a case:

```xml
<testsuites>
    <testsuite name="feature 1">
        <directory>some-dir</directory>
    </testsuite>
    <testsuite name="feature 2">
        <file>feature2-1Test.php</file>
        <file>feature2-2Test.php</file>
    </testsuite>
    <testsuite name="all">
        <testsuite>feature 1</testsuite>
        <testsuite>feature 2</testsuite>
    </testsuite>
</testsuites>
```

in this case paratest now would fetch all testfiles from feature 1 + 2 when executing.

paratest --testsuite=all

the implementation is recursive so n-level nesting should be possible.

I will also schedule a pull request for phpunit to support this (but this take some more time...^^)

### DIFF
--- a/src/ParaTest/Console/Testers/PHPUnit.php
+++ b/src/ParaTest/Console/Testers/PHPUnit.php
@@ -66,6 +66,17 @@ class PHPUnit extends Tester
         if ($input->getOption('runner') === 'WrapperRunner') {
             $runner = new WrapperRunner($this->getRunnerOptions($input));
         } else {
+            if ($input->getOption('runner') !== '') {
+                // because we want to have to bootstrap script inherited before check/initialization
+                $runnerOption = $this->getRunnerOptions($input);
+                $runnerClass = $input->getOption('runner');
+                if (class_exists($runnerClass)) {
+                    $runner = new $runnerClass($runnerOption);
+                }
+            }
+        }
+
+        if (!isset($runner)) {
             $runner = new Runner($this->getRunnerOptions($input));
         }
 
--- a/src/ParaTest/Runners/PHPUnit/Configuration.php
+++ b/src/ParaTest/Runners/PHPUnit/Configuration.php
@@ -1,5 +1,6 @@
 <?php
 namespace ParaTest\Runners\PHPUnit;
+use ParaTest\Parser\Parser;
 
 /**
  * Class Configuration
@@ -22,6 +23,8 @@ class Configuration
      * @var \SimpleXMLElement
      */
     protected $xml;
+
+    protected $availableNodes = array('file', 'directory', 'testsuite');
 
     /**
      * A collection of datastructures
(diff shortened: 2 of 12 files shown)


julianseeger:
Handy feature. Thanks for this!
We have it and phpunit don't. yay ^^

Also release 0.11.0

Slamdunk:
Hi, I'm writing https://github.com/paratestphp/paratest/pull/469 to check that all the XML used in this library tests validate against PHPUnit XSD schema.

The feature introduced here goes against PHPUnit current schema, and I think it's odd to support a schema non adherent to PHPUnit's one.

I'd like to remove this, tell me what you think please

julianseeger:
> Hi, I'm writing #469 to check that all the XML used in this library tests validate against PHPUnit XSD schema.
> 
> The feature introduced here goes against PHPUnit current schema, and I think it's odd to support a schema non adherent to PHPUnit's one.
> 
> I'd like to remove this, tell me what you think please

@Slamdunk whats the point of removing that feature?
The compatibility direction is paratest -> phpunit. Meaning that paratest should be usable as a drop-in-replacement for phpunit but that doesn't neccessarily mean that phpunit needs to work as a drop-in-replacement for paratest.

So this is just an addition if you want more features. Lika paratest is an addition if you want to run tests in parallel.

I see that it would have some kind of beauty to be fully compliant but is there any real reason to introduce such a backwards incompatible change?
And btw: if I understand the junit.xsd correctly, junit supports nested testsuites and as phpunit tries to be junits ugly brother I would suggest that you better place a pull request at phpunit to support nested testsuites ;) Just my 2c..

Slamdunk:
>  The compatibility direction is paratest -> phpunit. Meaning that paratest should be usable as a drop-in-replacement for phpunit but that doesn't neccessarily mean that phpunit needs to work as a drop-in-replacement for paratest.

Nowhere in the REAMDE I can read the `replacement` keyword.

My daily experience is that `paratest` is an awesome tool to run all the testsuite, even in CI, but for developing and debugging single tests/test-methods `phpunit` is still often needed.

> I see that it would have some kind of beauty to be fully compliant but is there any real reason to introduce such a backwards incompatible change?

If you, like me, run both `paratest` and `phpunit` regularly, using an unsupported configuration like the nested one gives you a warning:

```
vendor/bin/phpunit -c phpunit-files-dirs-mix-nested.xml
PHPUnit 9.2.6 by Sebastian Bergmann and contributors.

  Warning - The configuration file did not pass validation!
  The following problems have been detected:

  Line 17:
  - Element 'testsuite': This element is not expected. Expected is ( exclude ).

  Test results may not be as expected.


F.F.F.F..E.F..F......................                             37 / 37 (100%)
```

This warning is also raised during the `paratest` execution, and the drift in PHP community and tools like PHPUnit is to be more strict and blocking time by time; in the near future PHPUnit can decide to stop running against invalid configurations - and to be honest I hope it.

>  And btw: if I understand the junit.xsd correctly, junit supports nested testsuites and as phpunit tries to be junits ugly brother I would suggest that you better place a pull request at phpunit to support nested testsuites ;) Just my 2c..

I don't know junit.xsd, the validation I introduced in #469 run against [PHPUnit XSD](https://github.com/sebastianbergmann/phpunit/blob/master/phpunit.xsd).

I agree with you it would be nice PHPUnit to support nested testsuites as well, but for any feature the correct path for new ones should be PHPUnit -> Paratest, not the other way around. @chbiel seems did not followed up https://github.com/sebastianbergmann/phpunit/issues/1610 unfortunately.

> whats the point of removing that feature?

So here I reply to the first question you answered: the strict ParaTest sticks to PHPUnit, the lighter the maintenance cost is kept. In https://github.com/paratestphp/paratest/pull/451 @andreasschroth agreed with me to follow a simpler path for us to maintain ParaTest, and removing features PHPUnit doesn't support goes in this direction.

julianseeger:
If it is your project, I would suggest you just stop using that feature in your phpunit.xml. That way you wouldn't have a warning and everyone who uses it would still have that feature.  
But I don't use php anymore and have just been kicked out as a maintainer as far as I can tell so... yea just my thoughts. Maybe we can save some LOC.

andreasschroth:
I also don't really see the point to support features in the config that PhpUnit isn't supporting itself. I mean, the purpose of Paratest is to run the tests in parallel, that's the value of the tool, not to simplify the configuration.

Moreover, I'd be more than surprised if more than 0.x% something of Paratest users make use of this feature. So imo, can be dropped. But obviously it's a BC break, so it needs to be a new major release.

On the other hand, removing a feature "just for fun" without any good reason, makes no sense either. Do you think the code for this feature is that problematic @Slamdunk ?

Slamdunk:
The code itself is easy, keeping the knowledge clear in our mind for years and among many possible new maintainers/contributors is hard.

I'd love to actively maintain this library for years: for this target a lean tool-goal and clear code is essential.

> So imo, can be dropped. But obviously it's a BC break, so it needs to be a new major release.

With the actual semver adoption, that's not an issue to me if well documented in an upgrade-path file.

I have to say I plan to release a major release anyway for internal refactoring to get rid of some magics like [\ParaTest\Logging\MetaProvider::__call](https://github.com/paratestphp/paratest/blob/ce9268f3ae43e76a52ea0be9f5b80a75fbe80803/src/Logging/MetaProvider.php#L51) (that will be a review-requested PR)

andreasschroth:
Yeah, no objections to releasing a new major version. 👍 

Just wanted to mention it. :-)

julianseeger:
> I'd love to actively maintain this library for years: for this target a lean tool-goal and clear code is essential.

With that in mind the change makes definitely more sense. Previously I didn't understand why you don't just stop using that feature ;)

I would suggest to just go ahead with removing that feature and after the release anyone who uses it can start complaining. If nobody complains, fine. If too many people complain, the code is still in the git history to be brought back in with a hotfix.

andreasschroth:
OT: @julianseeger Maybe you can send me an e-mail (see my Github profile) as I don't have yours... would be glad if you're interested in maintaining this library again as well, I had to remove rights from previous maintainers as there was no contact at all anymore and also no communication.

Slamdunk:
Having @julianseeger back as maintainer would be awesome :+1:

julianseeger:
> Having @julianseeger back as maintainer would be awesome 👍

Won't happen, sorry. I moved to java/kotlin/python/julia and didn't touch PHP since version 7 🙈 
I tried to keep this library alive while I wasn't even using it anymore so I'm glad that it's now in your hands and you are passionate about maintaining/developing paratest.  
Don't hesitate to mention me if there is anything I can do for you ;)

Slamdunk:
Sticking to PHPUnit let us do things like https://github.com/paratestphp/paratest/pull/496 where we can drop every wheel we reinvented in favor of using PHPUnit ones.

Even if the PHPUnit used classes are internal, we skip pretty every bug related to out one on the topic, easing a lot the maintenance cost.

PR #151, opened 2015-02-08
Merged #65: Avoid simplexml_load_file warning when config path is a directory

julianseeger:
This fixes #65  (is basically a merge of that old PR).
I don't know what the old discussion was about (what diffs that discussion was about) but to me, it looks well tested.
A directory is passed in as configuration an both, phpunit.xml or phpunit.xml.dist are loaded of they exist. The only thing that could be added is a test for the load order (phpunit.xml.dist only if phpunit.xml does not exist). But I don't know what exactly @giorgiosironi was complaining about back then.

### DIFF
--- a/src/ParaTest/Runners/PHPUnit/Options.php
+++ b/src/ParaTest/Runners/PHPUnit/Options.php
@@ -192,19 +192,28 @@ class Options
      */
     protected function getConfigurationPath($filtered)
     {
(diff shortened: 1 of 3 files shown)


brianium:
@julianseeger looks good! Its cool to see these general improvements coming in. Let's keep it up!

Issue #60, opened 2013-07-24
Incomplete/Skipped Tests are ignored


dbaltas:
Incomplete/Skipped tests are not accounted for.
For example if you have 450 regular tests and 50 incomplete or skipped for a total of 500, you are expected to see the progress bar going 63/500, 126/500, where the 126 are the regular ones only.
So the progress bar will stop before 100% at 450 and in the end you will see 
OK (450 tests, X assertions)

sormy:
this problem fixed in the master

julianseeger:
Seems like there is a release missing. ... -> 0.12.2

TysonAndre:
PHPUnit 6 adds the test names for `skipped` and `incomplete` to the XML output. However, paratest doesn't yet, even in the planned 1.x/0.16 release supporting phpunit 6.

A followup task for https://github.com/brianium/paratest/pull/237  would be to output (in CLI mode) the classes and names of tests that were skipped

Slamdunk:
To me this seems fixed in the latest version, currently `4.1.0`

TysonAndre:
Same - I'm no longer seeing this issue in newer phpunit versions, and phpunit 6 support was dropped long ago. Totals are the same with phpunit/paratest

PR #154, opened 2015-02-13
@dataProvider tests parallelization in functional mode

sormy:
* @dataProvider support
  * --filter support (functional mode only)
  * --max-batch-size support (functional mode only)
  * print progress as vanilla phpunit
  * print progress without blank lines on windows default console
  * dump process command line for debug purposes if process do not return any results
  * one failed vanilla test: PHPUnitTest::testStopOnFailurePreventsStartingFurtherTestsAfterFailure
  * no auto tests for new features for now


sormy:
Just to note, one failed test marked as skipped because it seems like problem on test side (fix test) and not in code. On failed test is related to stop-on-failure feature. Previously it stops after method execution, now it stops based on batch execution.

julianseeger:
Phew, thanks a lot. It's going to take some time to review this because this pr includes a couple of changes at once.

You added a lot of code but I don't see any new tests, which is probably a bad sign. Would be nice if you could add some ... or a lot... . Maybe some tests could help to debug it because at least the output contains a bug:

I took a testsuite that looked like this:
```
julian@x1:~/PhpstormProjects/Metrics$ ./vendor/bin/paratest -f

Running phpunit in 5 processes with /home/julian/PhpstormProjects/Metrics/vendor/bin/phpunit. Functional mode is on

Configuration read from /home/julian/PhpstormProjects/Metrics/phpunit.xml

.........................................

Time: 630 ms, Memory: 6.25Mb

OK (41 tests, 134 assertions)
```

Then I added a dataProvider with 200 data sets to one of the methods which looks like this in **phpunit**:
```
julian@x1:~/PhpstormProjects/Metrics$ ./vendor/bin/phpunit 
PHPUnit 4.7-g7c1de6a by Sebastian Bergmann and contributors.

Configuration read from /home/julian/PhpstormProjects/Metrics/phpunit.xml

Deprecated configuration setting "strict" used

...............................................................  63 / 240 ( 26%)
............................................................... 126 / 240 ( 52%)
............................................................... 189 / 240 ( 78%)
...................................................

Time: 465 ms, Memory: 7.25Mb

OK (240 tests, 333 assertions)
```

And the current **brianium/paratest** master looks like this:
```
julian@x1:~/PhpstormProjects/Metrics$ ./vendor/bin/paratest -f

Running phpunit in 5 processes with /home/julian/PhpstormProjects/Metrics/vendor/bin/phpunit. Functional mode is on

Configuration read from /home/julian/PhpstormProjects/Metrics/phpunit.xml

................................................................. 65 ( 27%)
................................................................. 130 ( 54%)
................................................................. 195 ( 81%)
.............................................

Time: 1.39 seconds, Memory: 7.00Mb

OK (240 tests, 333 assertions)
```
Not exactly the same but consistent. (it looks the same if you omit the ```-f```).

Then I updated to **sormy/paratest** and it looked like this:
```
julian@x1:~/PhpstormProjects/Metrics$ ./vendor/bin/paratest -f

Running phpunit in 5 processes with /home/julian/PhpstormProjects/Metrics/vendor/bin/phpunit. Functional mode is on

Configuration read from /home/julian/PhpstormProjects/Metrics/phpunit.xml

...............................................................  63 / 240 ( 26%)
............................................................... 126 / 240 ( 52%)
............................................................... 189 / 393 ( 48%)
............................................................... 252 / 393 ( 64%)
............................................................... 315 / 393 ( 80%)
............................................................... 378 / 393 ( 96%)
...............

Time: 964 ms, Memory: 6.75Mb

OK (393 tests, 486 assertions)
```
There are 240 tests but it 393 tests are displayed. The total test count increases, too. When I omit ```-f```, then it looks exactly like phpunit.

I'm not yet sure what the max-batch-size does, but it has an influence on the output:
```
julian@x1:~/PhpstormProjects/Metrics$ ./vendor/bin/paratest -f --max-batch-size=1

Running phpunit in 5 processes with /home/julian/PhpstormProjects/Metrics/vendor/bin/phpunit. Functional mode is on

Configuration read from /home/julian/PhpstormProjects/Metrics/phpunit.xml

...............................................................  63 / 350 ( 18%)
............................................................... 126 / 360 ( 35%)
............................................................... 189 / 420 ( 45%)
............................................................... 252 / 480 ( 52%)
............................................................... 315 / 530 ( 59%)
............................................................... 378 / 530 ( 71%)
............................................................... 441 / 530 ( 83%)
............................................................... 504 / 530 ( 95%)
...........................

Time: 8.91 seconds, Memory: 12.00Mb

OK (531 tests, 625 assertions)
julian@x1:~/PhpstormProjects/Metrics$ ./vendor/bin/paratest -f --max-batch-size=5000

Running phpunit in 5 processes with /home/julian/PhpstormProjects/Metrics/vendor/bin/phpunit. Functional mode is on

Configuration read from /home/julian/PhpstormProjects/Metrics/phpunit.xml

...............................................................  63 / 240 ( 26%)
............................................................... 126 / 240 ( 52%)
............................................................... 189 / 240 ( 78%)
...................................................

Time: 765 ms, Memory: 6.50Mb

OK (240 tests, 333 assertions)
```
Would be nice if you could extend the readme somehow. Does it only apply to dataProviders?

sormy:
Thanks, it's just initial attempt to get feature work. We have suite which test DICOM stack with over 4000 tests which works for me exactly as phpunit and speedup is around 2x (18 min vs 38 min) on my macbook 2013.

If you can write fail test for your example it will be perfect!

Big max-batch-size can broke phpunit, because command line became too huge or regular expression became too hard to be parsed without changing php's pcre settings.

sormy:
Julian, it seems like new code have problems with dependent tests. Could you try run on end something without dependent tests?

julianseeger:
If you mean without an ```@depends``` annotation: I never use it. The original tests are from https://github.com/julianseeger/Metrics and there are no tests that depend on any other test.

sormy:
Ok, I found problem and fixed it. Also I added tests. Need you feedback before refactor code. max-batch-size affects on amount of atomic tests which will run in one isolated process - 1 atomic test is 1 test method without dataProvider or 1 dataset time for 1 method. Even if there are no datasets it will group different test() methods in batches - it also give boost on big amount of small tests because paratest will spent less time on spawning phpunit. I will add that note If you like that explanation.

sormy:
Also, I don't know what scrutinizer means: 1 failure condition was met by this inspection: issues.new.exists

sormy:
In my test scenario:
```
Artems-MBP:emsow sormy$ phpunit --filter Dicom
PHPUnit 4.5.0 by Sebastian Bergmann and contributors.

Configuration read from /Users/sormy/htdocs/emsow/phpunit.xml

.............................................................   61 / 5013 (  1%)
.............................................................  122 / 5013 (  2%)
.............................I...............................  183 / 5013 (  3%)
<cut>
............................................................. 4941 / 5013 ( 98%)
............................................................. 5002 / 5013 ( 99%)
..........I

Time: 41.03 minutes, Memory: 241.25Mb

OK, but incomplete, skipped, or risky tests!
Tests: 5013, Assertions: 83719, Incomplete: 2.
```

```
Artems-MBP:emsow sormy$ ../paratest/bin/paratest --filter Dicom

Running phpunit in 5 processes with /Users/sormy/htdocs/paratest/vendor/bin/phpunit. Functional mode is on

Configuration read from /Users/sormy/htdocs/emsow/phpunit.xml

.............................................................   61 / 5013 (  1%)
.............................................................  122 / 5013 (  2%)
<cut>
............................................................. 1098 / 5013 ( 21%)
............................................................. 1159 / 5013 ( 23%)
............................................................. 1220 / 5012 ( 24%)
............................................................. 1281 / 5012 ( 25%)
<cut>
............................................................. 4880 / 5012 ( 97%)
............................................................. 4941 / 5011 ( 98%)
............................................................. 5002 / 5011 ( 99%)
.........

Time: 17.43 minutes, Memory: 20.25Mb

OK (5011 tests, 83719 assertions)
```

Difference related to 2 incomplete tests and speedup is 2.35x on mobile core i7 with 2 cores

julianseeger:
About scrutinizer: take a look here https://scrutinizer-ci.com/g/brianium/paratest/inspections/91f9f970-f009-404d-a5c6-8114eb97fb4a and you see that the code quality of a couple of classes has decreased (but that didn't cause the failure) and that you introduced 82 new issues. An issue is something like a PSR violation. And PSR2 is the coding standard we stick to (https://github.com/php-fig/fig-standards/blob/master/accepted/PSR-2-coding-style-guide.md PSR2 implicitly includes PSR1, which includes PSR0).

Unfortunately, these new issues doesn't actually seem to be introduced by you. Scrutinizer shows
```
Analysis Tools Updated
Some analysis tools were updated to newer versions. This might result in changes which are unrelated to your code changes!
PHP Analyzer: 1.6.0 -> 1.6.1 (changelog)
```
and I can't find any issue that is ~11h old. The PHPAnalyzer Changelog has nothing important in it but I guess that scrutinizer marks every issue as new because of the new analyzer version. **So I will merge this PR even if scrutinizer screams** about these new issues. The classes that got worse can be beautified with a small refactoring which you already mentioned.

I'm sorry to say that you have to wait until tomorrow for feedback from me on your new code because I am handicapped today. But is looks good. Meanwhile you could take a look at the testStopOnFailurePreventsStartingFurtherTestsAfterFailure. Maybe it is enough to add a `max-batch-size` of `1` or make 1 the default for maximum BC. Whether or not the test still makes sense, it represents a feature that we don't want to loose. If one test has a failure and the `--stop-on-failure` flag is used, no new tests should be scheduled. With `-p 1` and a `max-batch-size` of 1 that means not a single test should run after the failing one. And every feature we don't want to loose needs a test ;)

sormy:
No problem,will wait for full feedback.

src/ParaTest/Runners/PHPUnit/TestMethod.php
```diff
@@ -20,17 +20,16 @@ class TestMethod extends ExecutableTest
...
-    protected $name;
+    protected $filters;
 
-    public function __construct($suitePath, $name)
+    public function __construct($testPath, $filters)
     {
-        $this->path = $suitePath;
-        $this->name = $name;
+        $this->path = $testPath;
+        $this->filters = (array)$filters;
```
julianseeger: what is this cast to array for?

src/ParaTest/Runners/PHPUnit/TestMethod.php
```diff
@@ -66,6 +78,6 @@ protected function prepareOptions($options)
      */
     public function getTestMethodCount()
     {
-        return 1;
+        return count($this->filters);
```
julianseeger: untested?

src/ParaTest/Runners/PHPUnit/SuiteLoader.php
```diff
@@ -214,29 +214,96 @@ private function executableTests($path, $classMethods)
...
      */
-    private function getMethodBatches($classMethods)
+    private function getMethodBatches($path, $class)
     {
+        $classMethods = $class->getMethods($this->options ? $this->options->annotations : array());
+        $maxBatchSize = $this->options && $this->options->functional ? $this->options->maxBatchSize : 0;
         $methodBatches = array();
         foreach ($classMethods as $method) {
+            $tests = $this->getMethodTests($path, $class, $method);
+            if (!$tests) {
```
julianseeger: Is there a usecase for this? If so, a test would be nice. If not, this code can be deleted.

src/ParaTest/Runners/PHPUnit/SuiteLoader.php
```diff
@@ -214,29 +214,96 @@ private function executableTests($path, $classMethods)
...
         }
         return $methodBatches;
     }
 
+    private function getMethodTests($path, $class, $method)
+    {
+        $result = array();
+
+        $dataProvider = $this->methodDataProvider($method);
+        if ($dataProvider) {
```
julianseeger: Would be nice if this branch could be covered on unit level. It's covered on functional level but based on the complexity of these methods, a unittest would fit in here very well.

src/ParaTest/Runners/PHPUnit/ResultPrinter.php
```diff
@@ -321,15 +338,14 @@ protected function getDefects(array $defects, $type)
...
-    protected function printProgress()
+    protected function getProgress()
     {
-        printf(
-            ' %' . $this->numTestsWidth . 'd (%3s%%)',
+        return sprintf(
+            ' %' . $this->numTestsWidth . 'd / %' . $this->numTestsWidth . 'd (%3s%%)',
             $this->casesProcessed,
-            floor(($this->casesProcessed / $this->totalCases) * 100)
+            $this->totalCases,
```
julianseeger: This is still not reliable.
Given you have skipped / incomplete tests, the total test count decreases. Example:
```
...............................................................  63 / 179 ( 35%)
............................................................... 126 / 136 ( 92%)
..........
```

The total count was originally removed because no total count is better than an incorrect total count.

test/functional/PHPUnitTest.php
```diff
@@ -238,6 +238,8 @@ public function testRunWithFatalRuntimeErrorWithTheWrapperRunnerOutputsError()
 
     public function testStopOnFailurePreventsStartingFurtherTestsAfterFailure()
     {
+        $this->markTestSkipped("Didn't work after @dataProvider related fixes");
```
julianseeger: add `'max-batch-size' => '1'` and you can remove the skip

test/unit/ParaTest/Console/Commands/ParaTestCommandTest.php
```diff
@@ -47,7 +47,9 @@ public function testConfiguredDefinitionWithPHPUnitTester()
             new InputOption('coverage-clover', null, InputOption::VALUE_REQUIRED, 'Generate code coverage report in Clover XML format.'),
             new InputOption('coverage-html', null, InputOption::VALUE_REQUIRED, 'Generate code coverage report in HTML format.'),
             new InputOption('coverage-php', null, InputOption::VALUE_REQUIRED, 'Serialize PHP_CodeCoverage object to file.'),
-            new InputOption('testsuite', null, InputOption::VALUE_OPTIONAL, 'Filter which testsuite to run')
+            new InputOption('testsuite', null, InputOption::VALUE_OPTIONAL, 'Filter which testsuite to run'),
+            new InputOption('max-batch-size', 'm', InputOption::VALUE_REQUIRED, 'Max batch size (only for functional mode).', 50),
```
julianseeger: max-batch-size=1 as a default is preferrable for BC, isn't it?
Especially when the Readme contains information about performance optimization and max-batch-size.

src/ParaTest/Runners/PHPUnit/TestMethod.php
```diff
@@ -38,9 +37,19 @@ public function __construct($suitePath, $name)
      *
      * @return string
      */
+    public function getFilters()
```
julianseeger: the docblock doesn't fit getFilters, but getName

test/unit/ParaTest/Console/Commands/ParaTestCommandTest.php
sormy: What is BC?

src/ParaTest/Runners/PHPUnit/ResultPrinter.php
sormy: I can remove code related to total count correction and write also S in progress for incomplete/skipped tests. But it is related to bug with wrong rendering of incomplete/skipped tests  =)

src/ParaTest/Runners/PHPUnit/SuiteLoader.php
sormy: Yes, if filter was used this method can return 0

src/ParaTest/Runners/PHPUnit/TestMethod.php
sormy: Also I'm thinking about renaming of this class because actually it's a set of tests for test file controlled by filter and filter can match with different methods and different datasets.

julianseeger:
**note on scrutinizer:**
On #157 , scrutinizer was disabled temporarily, so further commits should pass on scrutinizer and give more senseful feedback.

**feedback:**
Overall, it looks fine with the tests. Maybe you should consider writing more test on unit level, but this is easier with tdd then when writing them afterwards and especially hard on this project. But it is generally worth it because tests on unit level are more stable.

**stuff that should be fixed:**
* re-enable the skipped tests in the PHPUnitTest.php
* remove the total count from the ResultPrinter.php (not possible without reported skipped/incompletes)
* and of course refactoring is needed. Especially on SuiteLoader::getMethodBatches

**stuff to be discussed:**
* probably the default for max-batch-size should be 1. This is the closest to the current behavior.
It is not even guaranteed that a batch-size>1 is faster because the slow tests are not distributed equally between the batches, so a big batch full of slow tests may be slower than a couple of small, slow batches.

Again, thanks for your effort! This a much appreciated and often requested feature. But this process is neccessary to keep some quality standard for the project. And if you think what I'm saying is bs, I'm always to constructive feedback ;)

src/ParaTest/Runners/PHPUnit/TestMethod.php
sormy: for compatibility with other code (tests), which can pass string (one filter) instead of array.

src/ParaTest/Runners/PHPUnit/SuiteLoader.php
julianseeger: Is there a way to test this? If someone in the future removes this line (by not knowing what it is for or by accident) nobody will recognize.

src/ParaTest/Runners/PHPUnit/ResultPrinter.php
julianseeger: Interesting option. Is it save to say that every tests that doesn't show up in the junit-logs are skipped or incomplete? That way we could even print S for incompletes or skipped tests and fix #60 ...

test/unit/ParaTest/Console/Commands/ParaTestCommandTest.php
julianseeger: backwards compatibility
The current behavior is -f => 1 method per process/batch.

src/ParaTest/Runners/PHPUnit/SuiteLoader.php
sormy: If this code looks strange for you then it seems like I need to add not only tests but also a comment about that ;-)

test/unit/ParaTest/Console/Commands/ParaTestCommandTest.php
sormy: One test method can have more then one test case. So, actually we have three modes: parallel by test file, parallel by test method, and parallel by test case. My implementation replaces test method parallel with test case parallel. Even if we specify max-batch-size=1 new code will not work the same as old for methods with @dataProvider. I don't know if we actually need all of that methods =) Parallel by test case works the most effective for my needs. Also I found that testsuite grouping can be refactored to be activated on phase when engine group tests in batches - it will reduce amount of code. Current implementation has different logics (and different classes) for two modes: functional and regular.

src/ParaTest/Runners/PHPUnit/ResultPrinter.php
sormy: Also, I need to note that old version (before my fixes) even if it's not show total amount still show wrong percent, so from my point of vision showing self-corrected total amount of tests is not so bad because progress percent is self corrected too.

sormy:
> probably the default for max-batch-size should be 1.
default max-batch-size set to 0 and paratest works in that case the same as before my modifications
> re-enable the skipped tests in the PHPUnitTest.php
done
> remove the total count from the ResultPrinter.php (not possible without reported skipped/incompletes)
keep, implemented skipped/incomplete reporting as skipped for situations where group/group-exclude options are not passed to paratest
> and of course refactoring is needed. Especially on SuiteLoader::getMethodBatches
did some work to make code better
> ...and the docblocks at the TestMethod.php
done
> Overall, it looks fine with the tests. Maybe you should consider writing more test on unit level
only functional tests at least for now

waiting for your feedback

sormy:
new code base produce the following result in my test scenario, you can compare with my previous comment:

```
Artems-MBP:project sormy$ ../paratest/bin/paratest --filter Dicom --max-batch-size 100

Running phpunit in 5 processes with /Users/sormy/htdocs/paratest/vendor/bin/phpunit. Functional mode is ON.

Configuration read from /Users/sormy/htdocs/project/phpunit.xml

.............................................................   61 / 5013 (  1%)
.............................................................  122 / 5013 (  2%)
<skip>
............................................................. 2074 / 5013 ( 41%)
............................................................. 2135 / 5013 ( 42%)
......................S...................................... 2196 / 5013 ( 43%)
............................................................. 2257 / 5013 ( 45%)
<skip>
............................................................. 4697 / 5013 ( 93%)
............................................................. 4758 / 5013 ( 94%)
....................................S........................ 4819 / 5013 ( 96%)
............................................................. 4880 / 5013 ( 97%)
............................................................. 4941 / 5013 ( 98%)
............................................................. 5002 / 5013 ( 99%)
...........

Time: 18.54 minutes, Memory: 19.75Mb

OK, but incomplete, skipped, or risky tests!
Tests: 5013, Assertions: 83719, Incomplete: 2.
```

src/ParaTest/Runners/PHPUnit/ResultPrinter.php
sormy: > Is it save to say that every tests that doesn't show up in the junit-logs are skipped or incomplete?

almost safe, if code which calculate expected amount of tests is valid.

julianseeger:
To wrap this up here, I understand if you don't have time to close the last gaps. If you leave it at this state, I will of course merge it anyways but would fix the last "comments" myself before releasing it.
Thanks again for this massive +777-141

sormy:
I don't know project code rules, so I'm coding based on my experience. 
If you think that it's better to remove refactoring todo notes, it's ok - I will remove them. It's not a big question. 
Just would like to hear if I can do something else to make my part of work better without going deep into sources and rewriting some strange parts. 
My current list:
- remove todos
- replace @ to isset
- unit tests?
- more functional tests? is there any other functional tests i can code to test?
Let's merge that when you will have no any unresolved questions regarding my code.
Also I understand that code review requires time too, so It's your time too.

julianseeger:
I'm just trying to keep the contribution experience as enjoyable as possible while keeping an eye on the quality, which is - to some extend - subjective. About the todos: most pull-requests at paratest are about features people want/need or changes they have to make in order to do subsequent pull-requests (which then are related to their features). Thats probably normal for an open source project but as a result, not many people will address any todo they stumble upon. As a result, if it is possible to create a consistent solution with reasonable effort, I would always prefer such a solution over a todo. If it is not possible with reasonable effort, then let's leave the todo where it is. It always leaves the question "who is going to address it?". (With consistent I mean: because group and group-exclude are commented out, they need the `@` suppression, which others don't need).

It's not that TODOs are always a bad thing. The todo for the maximum command size for example is helpful because it gives an hint on how to implement this (`echo | xargs --show-limits` etc). But if you take for example TODOs in this style:
`//TODO some tests didn't work without/without the following line`
this should be true for every line in well tested code. Take well tested code, remove any sign and either some tests are failing or the code is not tested very well. So it seems pretty redundant to me. Maybe it would make sense to collapse all TODOs concerning `group` and `group-exclude` to a single todo and add it to the docblock of the defaults() method (or something like that).

So for your list:
 - [ ] remove/collapse todos (that don't contain more information than "tests are failing if/if not")
 - [ ] replace @ to isset
 - [x] functional tests: I think you covered all important paths on functional level. If you can think of any others, feel free to add them but, I don't think there is a need for more
 - [x] unit tests: looks like you covered all of your code. hard to tell because the coverage-reports seem to not recognize the code covered by subprocesses (spawned by the functional tests) but I don't think it adds much value to add additional unit tests for already logically covered code. If you want, you can take a look at https://scrutinizer-ci.com/g/brianium/paratest/inspections/60bbd396-9902-4816-96cf-e67c9f3532ce/code-structure/test-coverage which shows how the coverage changes with your diff. Like I said, it doesn't work very well with the functional tests but it says much about the unit tests.

sormy:
julianseeger, you are right regarding useless comments, I need either make them more addressed to problem or remove them if everything works fine without them.

sormy:
I would like to find a time to try to fix group, group-exclude and skipped/incomplete print support or leave to in right place

julianseeger:
Any progress here?

sormy:
Sorry Julian. I think I will have time on Saturday to complete everything and prepare for merge.

sormy:
Ok, Julian, I think, I'm done =) Could you please check if everything is ok.
PS: Travis lies =)

julianseeger:
I didn't go through all >1k loc of the change but it looked good before and the open points are addressed.
Thanks :)

Oh and there will be a separate issue about the PHP5.3 support. Thanks for addressing the compatibility issues, I hope we can get rid of that soon.

chbiel:
@sormy @julianseeger does this also address the issue (without -f option), that classes with dataproviders does not output correct % numbers of passed tests?

at the moment I see my test repeatly come from 79% to 99% and back to 79% when using @dataProvider

If not I will try to debug this on my side but feedback at first would be nice :)

sormy:
In non-functional mode it will work the same as before. Actually, I think that if we add a different types of parallelization (per test method, per test case, per test file, per test suite) for functional mode then we can  completely remove non-functional mode and get good percentage during test execution. Paratest in non-functional mode do not try to look into tests and estimate amount of tests and it's the reason why it works like it works.

julianseeger:
@chbiel yes. With functional mode and max-batch-size > 0, this feature is activated and progress is tracked for methods*dataproviders. So there should be no incorrect progress because of dataProviders.

@sormy In most cases, per test file and per test case should be the same and thats what non-functional mode does. per test method is functional-mode + max-batch-size=0. per dataProvider is functional + max-batch-size > 0. And per suite does not exist because it is so high level, that you can probably achieve this more easily by using per testcase concurrency or just use your build-tool to execute multiple commands concurrently.

Maybe we should always use this new feature to get always correct progress and split parsing / execution.

Your input is welcome on what the default should be and what options/combinations should be offered to make the use easier (and make the "optimizing speed" section of the readme intuitive). The current state is based on backwards compatibility. By default, nothing should have changed in the last releases. But when optimizing, you can turn of -f and --batch-max-size and look what does the job best. From my experience, you have like a 50% chance that data-provider-concurrency inceases or significally decreases performance (depends on thinks like bootstrapping and array-caches) and about a 80% chance to have a better performance in functional mode over non-functional. So there are no best options for everyone.

PR #156, opened 2015-02-15
Add seperat env for unique token

chbiel:
Fix for comment discussion in https://github.com/brianium/paratest/commit/ec7216b9cdbefa0d6c358913152d33117f01c23e


julianseeger:
Thanks.
A Token object would be nice to get rid of the array juggling.

PS: this is release 0.12.0

PR #157, opened 2015-02-18
uniqid() is not unique enough so add usleep to ensure uniqueness

chbiel:
When uniqid is executed to fast in a row the generated id's are unique. usleep ensures a uniqueid

would be 0.12.1 i think

### DIFF
--- a/.scrutinizer.yml
+++ b/.scrutinizer.yml
@@ -30,6 +30,3 @@ tools:
       standard: "PSR2"
   
   php_code_coverage: true
-  
-build_failure_conditions:
-  - 'issues.new.exists'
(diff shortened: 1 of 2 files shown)


chbiel:
okay :D there is something new with scrutinizer...

julianseeger:
**Scruzinizer:**
Ignore scrutinizer. You could remove the
```YAML
build_failure_conditions:
  - 'issues.new.exists
```
from the .scrutinizer.yml to make the build pass after a merge. We can turn it back on after those issues are addressed.

**uniqid:**
Ok, obvious bug. If the loop is faster than a ms, we get doubled results. But don't you get goosebumps when writing a usleep? ^^
`uniqid($i)` and `uniqid('', true)` should both fix the issue without a sleep because the token ($i) is unique within the process and the true should generate enough entropy to be unique within the same ms, shouldn't they?. (preferable prefix with $i because the $more_entropy flag generates uniqids with dots).

chbiel:
I tried ```uniqid('', true)``` but the generated number contains "." what I didnt find that beautiful. ´´´uniqid($i)´´´ did'nt came to my mind but is used now ;)

julianseeger:
>would be 0.12.1 i think

thanks :+1:

Issue #166, opened 2015-06-29
fix the build


julianseeger:
2 issues:
* hhvm-nightly no longer exists (has to be removed from the travis.yml)
* phpunit has corrected (?) it's output for warnings. They show up in the summary as tests, assertions and failures ... thus, the test https://github.com/brianium/paratest/blob/b107f324f01f1a5557589b946558a2a94e282c00/test/functional/PHPUnitTest.php#L311 has to be adjusted accordingly. (Looks strange that warnings show up as `FFWW`)

localheinz:
@julianseeger 

Here's a PR for removing the `hhvm-nightly` from the build matrix: https://github.com/brianium/paratest/pull/172.

julianseeger:
the issue related to warnings was propably caused by https://github.com/sebastianbergmann/phpunit/commit/821427232dd4028f8a13832219c4d73c73b3d879 which is a fix for https://github.com/sebastianbergmann/phpunit/issues/1317
phpunit 4.7.2 works but 4.7.3 does not

julianseeger:
fixed

PR #171, opened 2015-07-24
Enhancement: Enable caching of dependencies installed with Composer

localheinz:
This PR

* [x] enables caching of dependencies between builds


julianseeger:
Now that #166 is fixed, this can get in.
Thanks for your contributions

PR #173, opened 2015-08-23
Fix: Build

localheinz:
This PR

* [x] triggers a build to see if it even works


localheinz:
@brianium 

Current `master` build is failing, as you can see https://travis-ci.org/brianium/paratest/builds/76876779 - do you want to take a look?

julianseeger:
It's #166
PR welcome ;)

PR #177, opened 2015-09-01
avoid using simplexml_load_file for security reason

Djamiyo:
I've run into troubles trying to use paratest. For security reasons some hosters disable function simplexml_load_file, but our team went even further and just compiled php without it. So what do you think about this small change?

### DIFF
--- a/src/ParaTest/Runners/PHPUnit/Configuration.php
+++ b/src/ParaTest/Runners/PHPUnit/Configuration.php
@@ -38,9 +38,7 @@ class Configuration
     {
         $this->path = $path;
         if (file_exists($path)) {
-            $before = libxml_disable_entity_loader(false);
-            $this->xml = simplexml_load_file($path);
-            libxml_disable_entity_loader($before);
+            $this->xml = simplexml_load_string(file_get_contents($path));
         }
     }
 


Djamiyo:
Any thoughts?

brianium:
@Djamiyo this seems to address your issue. I don't see anything wrong with this code, and I am not sure if there are any ramifications. It looks good to me, but I would feel better if @julianseeger weighed in on it?

julianseeger:
I'm curious why this doesn't break the test of #90 ... `simplexml_load_string` shoud be affected by `libxml_disable_entity_loader` the same way `simplexml_load_file` is. And don't see a reason why any sane administrator should disable `simplexml_load_file` and leave `simplexml_load_string` active when it is vulnerable to XEE injections too.

Anyways, if this allows us to remove the code that enables the entity_loader and still works, I'm totally fine with that change ;)

julianseeger:
@djamiyo Sorry for the delay, I thought brianium would merge it after my comment.
I would like to merge your PR but that won't work with the latest changes to the composer.json

Djamiyo:
@julianseeger Sorry, I've changed this branch by mistake. The latest commit was reverted.

PR #181, opened 2015-09-28
FIX: SkippedOrIncompleteTest test fails if "S" in path

ThomasDelteil:
fix https://github.com/brianium/paratest/issues/178

Use a regex to capture the output instead of just counting the number of S


julianseeger:
thanks

Issue #19, opened 2013-01-30
Incorrent xml output if I use the dataProvider in tests


bayandin:
I think, this issue is continue #10.

If I use dataProvider in tests I get incorrent xml output (incorrect in the sence different from standart phpunit). The difference is that phpunit output has another `<testsuite>` tag and testsuite has a different attribute name. See example below.
- PHPUnit 3.7.13 from Composer,
- Paratest 0.4.1 or dev-master (545c4a7)

**Code example:**

``` php
<?php

class simpleTest extends PHPUnit_Framework_TestCase
{
    public function provider()
    {
        return array(
            array('foo'),
            array('bar'),
        );
    }

    /**
     * @dataProvider provider
     */
    public function test1()
    {
        $this->assertTrue(false);
    }
}
```

`vendor/bin/phpunit --log-junit=phpunit-junit.xml SimpleTest.php`

``` xml
<?xml version="1.0" encoding="UTF-8"?>
<testsuites>
  <testsuite name="SimpleTest" file="/home/a.bayandin/PhpstormProjects/Online4/tests/SimpleTest.php" tests="2" assertions="2" failures="2" errors="0" time="0.001455">
    <testsuite name="SimpleTest::test1" tests="2" assertions="2" failures="2" errors="0" time="0.001455">
      <testcase name="test1 with data set #0" assertions="1" time="0.001205">
        <failure type="PHPUnit_Framework_ExpectationFailedException">SimpleTest::test1 with data set #0 ('foo')
Failed asserting that false is true.

/home/a.bayandin/PhpstormProjects/Online4/tests/SimpleTest.php:18
</failure>
      </testcase>
      <testcase name="test1 with data set #1" assertions="1" time="0.000250">
        <failure type="PHPUnit_Framework_ExpectationFailedException">SimpleTest::test1 with data set #1 ('bar')
Failed asserting that false is true.

/home/a.bayandin/PhpstormProjects/Online4/tests/SimpleTest.php:18
</failure>
      </testcase>
    </testsuite>
  </testsuite>
</testsuites>
```

`vendor/bin/paratest -p 1 --phpunit=vendor/bin/phpunit --log-junit=paratest-junit.xml SimpleTest.php`

``` xml
<?xml version="1.0" encoding="UTF-8"?>
<testsuites>
  <testsuite name="" tests="2" assertions="2" failures="2" errors="0" time="0.001525" file="">
[...]

PR #185, opened 2015-11-12
Incorrent xml output if I use the dataProvider in tests

drefixs:
Incorrent xml output if I use the dataProvider in tests #19

### DIFF
--- a/src/ParaTest/Logging/LogInterpreter.php
+++ b/src/ParaTest/Logging/LogInterpreter.php
@@ -74,24 +74,25 @@ class LogInterpreter extends MetaProvider
             foreach ($reader->getSuites() as $suite) {
                 $cases = array_merge($cases, $suite->cases);
                 while (list( , $nested) = each($suite->suites)) {
-                    $this->extendEmptyCasesFromSuites($nested->cases, $suite);
+                    $this->extendEmptyCasesFromSuites($nested->cases, $suite, $nested);
                     $cases = array_merge($cases, $nested->cases);
                 }
             }
         }
         return $cases;
     }
-    
+
     /**
      * Fix problem with log from DataProvider
      *
      * @param array $cases
      * @param TestSuite $suite
+     * @param TestSuite $nestedSuite
      */
-    protected function extendEmptyCasesFromSuites($cases, TestSuite $suite)
+    protected function extendEmptyCasesFromSuites($cases, TestSuite $suite,TestSuite $nestedSuite)
     {
-        $class = !empty($suite->name)?$suite->name:"";
-        $file = !empty($suite->file)?$suite->file:"";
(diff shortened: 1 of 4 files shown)


julianseeger:
Thanks for the patch.
Could you please add a small test for this?

drefixs:
Ok if i have time then, i will add some fixture junit xml for dataProvider and make small test for that.

drefixs:
Pls review I did small test for this.
thx.
@julianseeger

julianseeger:
Thx. As the fix is referencing #19 which is pretty old, I would merge this PR.

But if you have some time left @drefixs there are some issues that could be fixed before the branch is merged:
1. even though the log makes much more sense than before (no empty names etc.), it still differs a lot from the phpunit log (compared to phpunit 5.0.9). I don't know how close we need to get, but at least a level of "testsuite" is missing
2. the tests executes every new line but it doesn't logically cover it
The lines [94](https://github.com/brianium/paratest/pull/185/files#diff-f9112ee9dade2252a9d4620293b76d50R94) and [95](https://github.com/brianium/paratest/pull/185/files#diff-f9112ee9dade2252a9d4620293b76d50R95) contain a lot of logic but the tests only checks for not-emptyness. Accordingly this would make the tests pass:
```PHP
$class = ".";
$file = ".";
```
So there is a test missing that checks if the values of $class and $file make sense.
And line 102 ```if (empty($case->file)) {``` can be removed without failing the test.

Feel free to fix these issues. ;)  It could be merged without further work, but #19 would be left open.

drefixs:
Thx, for review @julianseeger

1. Obtain https://raw.githubusercontent.com/windyroad/JUnit-Schema/master/JUnit.xsd PHPUnit performs invalid logs. testsuite haven't contain itself. It's Any matter to do log closer to PHPUnit format because it is invalid. The main target of this fix is to fix a critical  format requirement to make this log readable for report systems. There isn't enough data to do this absolutely valid to JUnit.xsd.

2. Thanks, these are right remarks, and I will carry out more complex unit tests

Thanks.

drefixs:
I decided to do logic more simple because it's bad idea to try get something from nested testsuite it doesn't aggregate any data from junit.xml log. I think to fix this problem directly there is 2-way. Wait for PHPUnit fix they JUnit log According to JUnit.xsd or rewrite paratest  Reader and Writer According to JUnit.xsd and add some adapter to Reader for invalid PHPUnit JUnit logs.

2. Done pls review :).

julianseeger:
Looks good. I'll check this tomorrow.

julianseeger:
Alright, thx for your additional fixes. (2) is fine now and with point (1) you are probably right. There is no hint anywhere that nested testcases are a valid occurence, not even in the phpunit docs as some kind of special case so I would also consider that a bug in phpunit. No point in adding buggy behavior just for compliance.

Issue #162, opened 2015-05-21
Doesn't work with PHP7


aik099:
Here is failed build using PHP 7: https://travis-ci.org/qa-tools/qa-tools/jobs/63512286

If I just run plain PHPUnit, then build passes sort of (ignore timeouts): https://travis-ci.org/qa-tools/qa-tools/jobs/63516774

Maybe it's failing because xdebug doesn't work on PHP 7.

aik099:
I think we should ignore coverage CLI options (or coverage phpunit.xml options) when no xdebug is installed. For example PHPUnit shows `The Xdebug extension is not loaded. No code coverage will be generated.` and continues to run tests.

The ParaTest is throwing an exception and does nothing:
```
 [PHP_CodeCoverage_Exception]       
  No code coverage driver available
```
In fact maybe internally PHPUnit catches that PHP_CodeCoverage exception and shows nice message instead.

julianseeger:
Makes sense for the default behavior. But additionally the tests would have to be skipped because they make no sense without xdebug.

aik099:
I hope, that that change would fix Travis CI PHP7 test run.

julianseeger:
The time works for us. It's still failing with PHP7 but now it looks fixable.
Probably related to the changed behavior of RuntimeErrors and Fatals (quick guess...)

https://travis-ci.org/brianium/paratest/jobs/77806863

julianseeger:
Help is appreciated here!
Next week, the final release of PHP7 is on schedule and the functional tests are still failing:
https://travis-ci.org/brianium/paratest/jobs/88874512

Would be nice if someone would find the time to dig in an find out why $proc->getErrorOutput() is empty and fatals result in an exitCode of 2 instead of 1. I'll have to disallow php7 builds to fail in the travis build-matrix on november 12th, which will result in a build-break for the whole build.

PR #191, opened 2015-12-05
PHP 7 & PHPUnit 5

### LINKED ISSUE #189: No coverage support in PHPUnit 5
```
Error:         No whitelist configured, no code coverage will be generated
```
https://travis-ci.org/brianium/paratest/jobs/94485558

PHPUnit 5 seems to require a whitelist configured when running with coverage.
The paratest build fails. Maybe it is only a problem with the phpunit test itself (but maybe not ^^)

julianseeger:
This adds fixes to support php 7 and phpunit 5. ( fixes #189 and #162 )
It is mainly build-/test-related:
* with php 7, phpunit catches fatal errors like normal errors => no special failure handling neccessary, they are skipped with
```PHP
if (PHP_VERSION_ID >= 70000) {
    $this->markTestSkipped('fatals are handled like normal exceptions with php7');
}
```
* phpunit 5 requires a whitelist to be defined when tracking coverage

the fix for this adds a deep dependency to phpunit, because paratest can only support the ```--whitelist``` option, when phpunit does, and it does only for versions >=5.0.0
```PHP
public static function isWhitelistSupported() {
    return Comparator::greaterThanOrEqualTo(\PHPUnit_Runner_Version::id(), '5.0.0');
}
```

So I need some feedback here, especially from @brianium (and anyone else interested) whether this is justifiable or we have better options.

PS: the current phpunit-dev-master does not work anymore so there must be a problem introduced after phpunit 5.1.0, but I couldn't track it down yet

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -16,6 +16,9 @@ matrix:
   fast_finish: true
   allow_failures:
     - env: PHPUNIT_VERSION='*@dev'
+  exclude:
+    - php: 5.5
+      env: PHPUNIT_VERSION='5.*@stable'
 
 env:
     - PHPUNIT_VERSION='3.7.*@stable'
--- a/.travis.yml
+++ b/.travis.yml
@@ -16,7 +16,6 @@ matrix:
   fast_finish: true
   allow_failures:
     - env: PHPUNIT_VERSION='*@dev'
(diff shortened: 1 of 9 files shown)


marco-jantke:
:+1:

PR #192, opened 2015-12-08
Support groups in phpunit.xml.

grkej:


julianseeger:
Ok, I assume ```e1dfc4bdaeb543f7929b768bfc085cc0dc857ea4``` to be an accident, because we won't change the projects name ;)
Additionally, your PR doesn't include any tests and the tests are failing (see https://travis-ci.org/brianium/paratest/builds/95685745 ). Please fix these issus to finish this PR, thanks.

julianseeger:
After two years, I don't expect to hear anything anymore. So thanks for the try, next time please add some tests.

Issue #180, opened 2015-09-03
paratest ignores the <exclude> tags from the phpunit config.xml


ThomasDelteil:
As per the title, paratest ignores the type of tags in the `<testsuite>` node
- `<exclude>`
- `<file>`
- `<directory>`

They are treated without distinction

rscafi:
:+1:, it was driving me crazy

ThomasDelteil:
I actually have a fix somewhere for that, I will see if I can get it in order and properly tested and issue a PR. As far as I remember I had to refactor a few classes to make it work properly though

ThomasDelteil:
@rscafi I have added tests and opened a PR. That would be helpful if you had time to checkout this branch and test it against your config see if it behaves correctly:
https://github.com/ThomasDelteil/paratest/tree/fix_exclude_folder

PR #195, opened 2016-01-21
[Issue 180] Fix exclude folder

ThomasDelteil:
Fix https://github.com/brianium/paratest/issues/180

I have split the SuiteLoader into SuiteLoader and TestFileLoader. This helps separating the concerns between the files and the suites.

A SuitePath now is aware of the folders excluded within the Suite.
The TestFileLoader makes use of this information when resolving the individual file paths for the suite.


julianseeger:
Thx. According to #180 I'll wait for feedback from @rscafi before reviewing it.

rscafi:
Thanks for the contribution! I can confirm that, at least for my use case, it's working properly.

PR #196, opened 2016-02-25
Bump version

MarkVaughn:
it's confusing thinking your dependencies are correct yet seeing a different version number


julianseeger:
Right. But the next version/tag would be 0.13.3 ;)
Can't we extract it from composer or something? Otherwise it will always be outdated...

MarkVaughn:
word

PR #198, opened 2016-02-25
Update SuitePath.php

MarkVaughn:
Get this Warning: preg_quote() expects parameter 1 to be string, array given in brianium/paratest/src/ParaTest/Runners/PHPUnit/SuitePath.php

Due to suffix being an empty array in the construtor


julianseeger:
The build is failing on your PR. Not your fault but would you mind having a look at it? Maybe it's quick to fix. Probably one of the latest versions of PHPUnit is acting different on warnings.

julianseeger:
@michaelbutler fixed the build, thanks for the contribution ;)

Issue #146, opened 2015-02-04
Using @group throws exception 


jameshd:
Whenever we use the @group annotation we encounter the following exception

```
[RuntimeException]                                                                                                                      
  Log file /tmp/PT_YQNOiy is empty.                                                                                                       
                  This means a PHPUnit process was unable to run /src/test/Components/CompetitionFilter/FilterTest.php  
                  Maybe there is more than one class in this file.
```

Having traced the code in ParaTest we've found where the exception is being thrown and that the Reader->hasResults() method is returning false, thus causing the exception. 

However, should this code even be being executed as the test in the Exception does not contain the requested @group annotation?

With the following lines commented out  - the tests work as expected, but there could be underlying issues to which we're not aware. 

``` php

       if (!$reader->hasResults()) {
            throw new \RuntimeException("Log file " . $test->getTempFile() . " is empty.
                This means a PHPUnit process was unable to run " . $test->getPath() . "
                Maybe there is more than one class in this file.");
        }
        $this->results->addReader($reader);
```

Any advice would be appreciated.

julianseeger:
Maybe you've hit a bug.

The codeblock that throws this exception is relatively new and the [integration test for the group-filter](https://github.com/brianium/paratest/blob/948105b7e98d1ce14f56139284d369f8237dce13/test/functional/GroupTest.php) seems to not cover this scenario where a whole file has no matching group attribute.
This can be confirmed easily by extending the fixture for the [GroupTest](https://github.com/brianium/paratest/blob/948105b7e98d1ce14f56139284d369f8237dce13/test/functional/GroupTest.php) with a second file that contains no matching groups. In your case, this should then lead to this exception.

gabrielpreston:
I run into the same issue when I tried upgrading from phpunit 3.7 to 4.0. It seems that starting with v4, the log file is written differently. In 3.7, the log file included information about the test and said there were zero tests run, whereas in 4.0, the log file terminates with <testsuites/> and includes no more information.

gabrielpreston:
This is how the junit log file looks now when using phpunit 4 and you pass a @group that doesn't match a test that gets passed from paratest to phpunit.
```
<?xml version="1.0" encoding="UTF-8"?>
<testsuites/>
```

fiddur:
I have the same issue with --exclude-group

matteomoretti:
Running paratest from dev-master with phpunit 4.6.6 the bug exists. It disappears if you comment out 
```php
[...]

Issue #164, opened 2015-06-19
Error with group option


nirav1911:
I am getting an error when I use --group with paratest and some of the test cases don't have any group defined in the tests.

Error: Maybe there is more than one class in this file.

sormy:
What version do you have?

jaredmcateer:
I have this problem as well. I'm using the latest dev-master via composer as of today. I tried to exclude a group and I am getting this error on a file that has the annotation:

     paratest -p4 --configuration phpunit.xml --exclude-group=synchronousTests Website/

    Running phpunit in 4 processes with /data/istock/vendor/bin/phpunit

    Configuration read from /data/istock/ZF/unittest/phpunit.xml
                                                                                                                                                                                                                                              
    [RuntimeException]                                                                                                                                                                                                                          
    The process: '/data/www/vendor/bin/phpunit' '--configuration' '/data/www/ZF/unittest/phpunit.xml' '--exclude-group' 'synchronousTests' '--log-junit' '/tmp/PT_dCj3MU' 'Website_Asset_NamespaceTest' 'Website/Asset/NamespaceTest.php'                                                                                                                                                                                                                                           
    Log file "/tmp/PT_dCj3MU" is empty.                                                                                                                                                                                                         
    This means a PHPUnit process was unable to run "Website/Asset/NamespaceTest.php"                                                                                                                                                            
    Maybe there is more than one class in this file.

sormy:
Could you write fail phpunit test for that case?

jaredmcateer:
Sure thing

    paratest -p4 --exclude-group=skip MyTest.php

Works:
```php
<?php
class MyTest extends PHPUnit_Framework_TestCase
{
    /**
     * @group skip
     */
    public function testGroup()
    {
        $this->assertTrue(true);
    }

    public function testGroup2()
    {
        $this->assertTrue(true);
    }
}
```

Fails:
```php
<?php
class MyTest extends PHPUnit_Framework_TestCase
{
    /**
     * @group skip
     */
    public function testGroup()
    {
        $this->assertTrue(true);
    }

    /**
     * @group skip
     */    
    public function testGroup2()
    {
        $this->assertTrue(true);
    }
}
```

Fails:
```php
<?php
/**
* @group skip
*/
class MyTest extends PHPUnit_Framework_TestCase
{

    public function testGroup()
    {
        $this->assertTrue(true);
    }

    public function testGroup2()
    {
        $this->assertTrue(true);
    }
}
```

[...]

PR #201, opened 2016-04-13
Support @group, @large, @small, @medium in test class and test method annotation

RojerUA:
Fix for bug: #164


julianseeger:
Thanks for the contribution. Could you please add a test for this?
And someone has to fix the testcompatibility to the latest phpunit version that travis can confirm the tests

src/ParaTest/Runners/PHPUnit/SuiteLoader.php
```diff
@@ -230,7 +236,7 @@ private function getMethodTests($class, $method, $useDataProvider = false)
 
     private function testMatchGroupOptions($groups)
     {
-        if (empty($groups)) {
+        if (empty($this->options->groups) && empty($this->options->excludeGroups) && empty($groups)) {
```
carlo1138: This breaks for test methods without annotation inside a test class with one.
```php
/**
 * @group foo
 */
class TestClass
{
    public function testMethod() {}
}
```
It works on `initSuites` but then it will fail for `getMethodTests`

carlo1138: Ping 😄

julianseeger: this PR needs work anyways. So either there are tests added and your issue is fixed @carlo1138 or this PR won't get merged.

oligriffiths:
Hi. Just wonder what the status is on this? We're encountering this issue https://github.com/brianium/paratest/issues/146

julianseeger:
Hi @oligriffiths . Status is unchanged: The new code is untested and as @carlo1138 stated, 
> This breaks for test methods without annotation inside a test class with one.

So unfortunately the effort in this PR seems wasted. If the state gets more clear by it, I could close this PR. Or someone (maybe you?) could fix it and add some tests. I guess it's easier to complete this PR than to come up with a new fix from scratch.

julianseeger:
Edit: there is a newer PR for #146: https://github.com/brianium/paratest/pull/225
It "just" waits for the master to be fixed, because merging on a broken master is a bad thing. Unfortunately I currently don't have the time to do it myself, so it's up to the world to fix it. Until then, no new PR will get in. You may have a look at travis to see what went wrong on the last build.

oligriffiths:
@julianseeger any reason this was closed?

julianseeger:
https://github.com/brianium/paratest/pull/225 is merged and should fix #146 (release coming soon).
Besides that, @RojerUA didn't leave a message in this PR for about nearly a year while this code breaks existing functionality AND is not tested. So .. if you are still missing functionality that #225 didn't fix, feel free to create a new PR with your code or a fixed and tested version of this code. I doubt that @RojerUA will add something here

PR #203, opened 2016-05-07
Fix some documentation issues found in Scrutinizer

michaelbutler:
Fixed some of the issues found in Scrutinizer:

* PHPDoc for read-only properties
* Defined protected properties for some dynamically defined properties.

### DIFF
--- a/src/ParaTest/Logging/LogInterpreter.php
+++ b/src/ParaTest/Logging/LogInterpreter.php
@@ -5,6 +5,12 @@ use ParaTest\Logging\JUnit\Reader;
 use ParaTest\Logging\JUnit\TestSuite;
 use ParaTest\Logging\MetaProvider;
 
+/**
+ * Class LogInterpreter
+ * @package ParaTest\Logging
+ * @method array getFailures
+ * @method array getErrors
+ */
 class LogInterpreter extends MetaProvider
 {
     /**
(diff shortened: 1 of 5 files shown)


julianseeger:
Thanks for the PR but as long as nobody fixes the build (=the tests for the current phpunit version), nothing will get merged :(

michaelbutler:
Totally agree, I'm going to attempt to figure out why it's failing and fix
it in a separate pr.  For starters, it passes on my laptop with the same
versions of everything.

julianseeger:
Now that build is green (or will be after a rebase/merge), most of this PR looks pretty redundant to me (especially the `@package` annotation and the `Class` DocBlock, I mean seriously ^^) but if it satisfies some IDEs and probably won't hurt anyone.

Just out of interest: which IDEs have problems with autocompletion of real methods when a magic method is present?

michaelbutler:
@julianseeger you're right about 'Class' and '@package', those really aren't needed so I'll take them out. Basically what I was going by was the Scrutinizer job. I think all that stuff can be customized maybe to turn certain features off. Still, for things like @property-read I think it does actually make sense. I'm using PHPStorm for the record. 
I also added `protected $variable` statements which I think is definitely the most important thing to add. It's bad practice in PHP to define variables on the fly like $this->foo inside a constructor or anywhere, without having an explicit private/protected/public definition in the class.

michaelbutler:
ok, no idea why the build is failing for mostly just comment PHPdoc changes. will look into it later.

michaelbutler:
Closing until further notice.

PR #204, opened 2016-05-08
Change unit test assertion depending on the version of PHPUnit being used

michaelbutler:
As described below in comments, PHPUnit 5.1+ changed the way test warnings are treated in test suites. This is not to be confused with E_WARNINGS that are raised during tests. These are things like "test method is marked as private" or "non-existent data provider specified". In PHPUnit 4.8 these were printed as warnings in the test output, and they also caused a failed build (exit code > 0). In PHPUnit 5.1+ these still output warnings, but they do NOT output a failed exit code.

This PR adds a PHPUnit version check to change the assertion based on what we actually expect to happen.

### DIFF
--- a/composer.json
+++ b/composer.json
@@ -2,7 +2,7 @@
     "name": "brianium/paratest",
     "require": {
         "php": ">=5.4.0",
-        "phpunit/phpunit": ">=3.7.8",
+        "phpunit/phpunit": "4.*@stable",
         "phpunit/php-timer": ">=1.0.4",
         "symfony/console": "~2.3|~3.0",
         "symfony/process": "~2.3|~3.0",
--- a/composer.json
+++ b/composer.json
@@ -2,7 +2,7 @@
     "name": "brianium/paratest",
     "require": {
         "php": ">=5.4.0",
-        "phpunit/phpunit": "4.*@stable",
+        "phpunit/phpunit": ">=3.7.8",
         "phpunit/php-timer": ">=1.0.4",
         "symfony/console": "~2.3|~3.0",
         "symfony/process": "~2.3|~3.0",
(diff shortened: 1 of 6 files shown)


michaelbutler:
@julianseeger @brianium I have confirmed that the paratest build is failing in PHPUnit 5.x because in PHPUnit 5.1+, (https://github.com/sebastianbergmann/phpunit/blob/5.1/ChangeLog-5.1.md) the way test warnings (not to be confused with E_WARNINGs emitted by actual code) are handled was changed. 

In PHPUnit 4, these test warnings, such as a testFooBar() method being private, would cause a failed test suite (exit code > 0). In PHPUnit 5.1+, this was changed to print a warning "W" but allow a regular exit code 0.

As a proposed solution, though this might be hacky, I could implement a PHP version specific test to test the different PHPUnit functionalities. PHPUnit 5.x is only compatible with PHP 5.6+, and PHP Unit 4.8- is only compatible with PHP 5.5 and below. Therefore we can have one test be <= PHP 5.5, and the other test be >= PHP 5.6. See https://phpunit.de/manual/current/en/appendixes.configuration.html#appendixes.configuration.testsuites

I'll go ahead and update this PR with the proposed solution.

michaelbutler:
ok guys let me know what you think. I basically added a PHPUnit version check in there and the build passes in all cases!
Not sure else how to handle this, besides just removing the test. PHPUnit is an external dependency here so if they change functionality we need to be able to deal with that.

julianseeger:
Nice one, thx.
I'm fine with the `if` in the test as the outcome it fixed per environment. Other solution would be to have two tests and skip one or the other, but then we would have this ugly `S` ;)

Travis is fine, scrutinizer is fine, I'm fine, merged!

Issue #199, opened 2016-03-18
XML output cannot be parsed by Jenkins


SvenRtbg:
When running a test suite, the resulting XML is correctly parsed by Jenkins when only using PHPUnit, but fails when using Paratest.

The build is triggered by a Phing build.xml

```
<property name="PARATEST_CALL" value="&quot;${project.basedir}/vendor/bin/paratest&quot;"/>
<property name="PARATEST_OPTIONS" value="-p 8"/>
<property name="PHPUNIT_CALL" value="&quot;${project.basedir}/vendor/bin/phpunit&quot;"/>

<exec command="cd tests/integration; ${PARATEST_CALL} ${PARATEST_OPTIONS} --phpunit ${PHPUNIT_CALL} --log-junit &quot;${BUILD_DIRECTORY}/results-integration.xml&quot;" passthru="true" checkreturn="true"/>
```

This fails.

However, when I fix the output file just a little bit, it works:

```
<exec command="cd tests/integration; ${PARATEST_CALL} ${PARATEST_OPTIONS} --phpunit ${PHPUNIT_CALL} --log-junit &quot;${BUILD_DIRECTORY}/results-integration.xml&quot;; ret=$?; sed -i -e 's/line=&quot;&quot;/line=&quot;0&quot;/' ${BUILD_DIRECTORY}/results-integration.xml; (exit $ret);" passthru="true" checkreturn="true"/>
```

The important part is the SED operation (the surrounding part is just saving the exit code):

sed -i -e 's/line=&quot;&quot;/line=&quot;0&quot;/'

Line numbers always have to be integers, they cannot be empty.

I'm using Paratest 0.13.2 and PHPUnit 4.8.23, but the problem has been bugging me for a long time (read: has been present in older versions of both softwares as well).

SvenRtbg:
The Jenkins xunit plugin has a schema for the result.xml files. See https://github.com/jenkinsci/xunit-plugin/blob/master/src/main/resources/org/jenkinsci/plugins/xunit/types/phpunit-2.0.xsd

mabu-github:
Another solution could be not to write out the "line" attribute because it is optional.
`<xs:attribute name="line" type="xs:integer" use="optional"/>`

SvenRtbg:
https://github.com/brianium/paratest/issues/19 looks related to this.

michaelbutler:
I hope that my PR fixes this problem. Please confirm if you can by checking out this branch and using it in jenkins.

[...]

PR #207, opened 2016-06-03
[wip] Fix tests error after phpunit/phpcov new release

Webysther:
composer show

```
brianium/habitat                       v1.0.0
brianium/paratest                      dev-fix-tests
classpreloader/classpreloader          3.0.0
composer/semver                        1.4.1
dnoegel/php-xdg-base-dir               0.1
doctrine/inflector                     v1.1.0
doctrine/instantiator                  1.0.5
folha/illuminate                       1.0.0-rc4
folha/validation                       1.0.0-rc6
hamcrest/hamcrest-php                  v1.2.2
jakub-onderka/php-console-color        0.1
jakub-onderka/php-console-highlighter  v0.3.2
jeremeamia/SuperClosure                2.2.0
laravel/framework                      v5.2.35
laravelcollective/html                 v5.2.4
league/flysystem                       1.0.24
mockery/mockery                        0.9.5
monolog/monolog                        1.19.0
mtdowling/cron-expression              v1.1.0
myclabs/deep-copy                      1.5.1
nesbot/carbon                          1.21.0
nikic/php-parser                       v2.1.0
paragonie/random_compat                v1.4.1
pdepend/pdepend                        2.2.4
phpdocumentor/reflection-docblock      2.0.4
phploc/phploc                          3.0.1
phpmd/phpmd                            2.4.3
phpspec/php-diff                       v1.0.2
phpspec/phpspec                        2.5.0
phpspec/prophecy                       v1.6.0
phpunit/php-code-coverage              4.0.0
phpunit/php-file-iterator              1.4.1
phpunit/php-text-template              1.2.1
phpunit/php-timer                      1.0.8
phpunit/php-token-stream               1.4.8
phpunit/phpcov                         3.1.0
phpunit/phpunit                        5.4.2
phpunit/phpunit-mock-objects           3.2.0
psr/log                                1.0.0
psy/psysh                              v0.7.2
respect/validation                     1.1.2
sebastian/code-unit-reverse-lookup     1.0.0
sebastian/comparator                   1.2.0
sebastian/diff                         1.4.1
sebastian/environment                  1.3.7
sebastian/exporter                     1.2.1
sebastian/finder-facade                1.2.1
sebastian/git                          2.1.2
sebastian/global-state                 1.1.1
sebastian/object-enumerator            1.0.0
sebastian/phpcpd                       2.0.4
sebastian/recursion-context            1.0.2
sebastian/resource-operations          1.0.0
sebastian/version                      2.0.0
squizlabs/php_codesniffer              2.6.1
swiftmailer/swiftmailer                v5.4.2
symfony/config                         v3.1.0
symfony/console                        v3.0.6
symfony/debug                          v3.0.6
symfony/dependency-injection           v3.1.0
symfony/event-dispatcher               v3.1.0
symfony/filesystem                     v3.1.0
symfony/finder                         v3.0.6
symfony/http-foundation                v3.0.6
symfony/http-kernel                    v3.0.6
symfony/polyfill-mbstring              v1.2.0
symfony/polyfill-php56                 v1.2.0
symfony/polyfill-util                  v1.2.0
symfony/process                        v3.0.6
symfony/routing                        v3.0.6
symfony/translation                    v3.0.6
symfony/var-dumper                     v3.0.6
symfony/yaml                           v3.1.0
theseer/fdomdocument                   1.6.1
vlucas/phpdotenv                       v2.2.1
```

### DIFF
--- a/README.md
+++ b/README.md
@@ -40,7 +40,7 @@ Options:
  --help (-h)           Display this help message.
  --coverage-clover     Generate code coverage report in Clover XML format.
  --coverage-html       Generate code coverage report in HTML format.
- --coverage-php        Serialize PHP_CodeCoverage object to file.
+ --coverage-php        Serialize \SebastianBergmann\CodeCoverage\CodeCoverage object to file.
  --max-batch-size (-m) Max batch size (only for functional mode). (default: 0)
  --filter              Filter (only for functional mode).
  --phpunit             The PHPUnit binary to execute. (default: vendor/bin/phpunit)
--- a/src/ParaTest/Console/Commands/ParaTestCommand.php
+++ b/src/ParaTest/Console/Commands/ParaTestCommand.php
@@ -41,7 +41,7 @@ class ParaTestCommand extends Command
             ->addOption('help', 'h', InputOption::VALUE_NONE, 'Display this help message.')
             ->addOption('coverage-clover', null, InputOption::VALUE_REQUIRED, 'Generate code coverage report in Clover XML format.')
             ->addOption('coverage-html', null, InputOption::VALUE_REQUIRED, 'Generate code coverage report in HTML format.')
-            ->addOption('coverage-php', null, InputOption::VALUE_REQUIRED, 'Serialize PHP_CodeCoverage object to file.')
+            ->addOption('coverage-php', null, InputOption::VALUE_REQUIRED, 'Serialize \SebastianBergmann\CodeCoverage\CodeCoverage object to file.')
             ->addOption('max-batch-size', 'm', InputOption::VALUE_REQUIRED, 'Max batch size (only for functional mode).', 0)
             ->addOption('filter', null, InputOption::VALUE_REQUIRED, 'Filter (only for functional mode).');
 
--- a/src/ParaTest/Coverage/CoverageMerger.php
+++ b/src/ParaTest/Coverage/CoverageMerger.php
@@ -2,16 +2,14 @@
 
 namespace ParaTest\Coverage;
 
-use \SebastianBergmann\CodeCoverage\CodeCoverage as CodeCoverage;
-
 class CoverageMerger
 {
     /**
-     * @var \CodeCoverage
+     * @var \SebastianBergmann\CodeCoverage\CodeCoverage
      */
     private $coverage = null;
 
-    public function addCoverage(\CodeCoverage $coverage)
+    public function addCoverage(\SebastianBergmann\CodeCoverage\CodeCoverage $coverage)
     {
         if ($this->coverage == null) {
             $this->setCoverage($coverage);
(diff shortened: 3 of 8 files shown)


julianseeger:
In case it make this PR easier: as far as I'm concerned, we can drop support for phpunit 3.* (in case @brianium agrees). It's not even listen any longer on phpunit.de assuimg that the official website.
But we should support phpunit 4.8 as it's supported by phpunit till `February 3, 2017`. So feel free to drop the `- PHPUNIT_VERSION='3.7.*@stable'` from the travis.yml.

Webysther:
Thanks! :)

aik099:
Would it still work with previous version of PHPCov?

Webysther:
@aik099 I try make compatible, but, i do no if is possible.

michaelbutler:
I don't agree with this change because now it means you cannot use paratest on older than PHPUnit 5.4.2. e.g. cannot use PHPUnit 4.8. I also had this problem and I fixed in in this PR: https://github.com/brianium/paratest/pull/208
By directly requiring `phpunit/php-code-coverage:3.*` we can 'lock' the paratest project to use that version of code coverage. The tradeoff is that you will not be able to use PHPUnit > 5.4.2 at all. 
This fix should be in place until we want to drop support for PHP 5.5 AND PHPUnit 4.8, at which point we can move to the new CodeCoverage namespace.

julianseeger:
With the current state of this PR, we would have to add `phpunit/php-code-coverage >= 4.0.0`. Then everyone using php-code-coverage would get an older release of phpunit and everyone using >=4.0.0 would get the next release of phpunit.

#208 on the opposite only allows the usage of php-code-coverage < 4 ... so this sounds a little bit better. But can't we just do a version switch? I cound 4 methods (ignoring type hints) where the code would have to switch the implementation based on available classes and the evaluation,
 which strategy to use could be made on startup I guess.

julianseeger:
To clarify: #208 is merged because it is definitely better than the previous master. I would still prefer supporting both versions of phpcc. But I agree with @michaelbutler that dropping phpcc 3.x support would probably (at least currently) hit most users. It would be hard to decide when a switch to >=4.0 is possible when we don't support 3.x anymore.

michaelbutler:
@julianseeger yes, ideally to support everything we have to do a version check and remove the typehints from the shared class -- which IMO is kind of crappy reducing to a loose interface without typehints, but a necessary evil. If PHP Code Coverage had implemented interfaces it would have actually solved this problem

Webysther:
@michaelbutler You have an excellent thinking about support for all users, but I disagree with the support model.

You do not need a bad code to work for both, as the support is at the end of PHP 5.5, users who need to older versions, use the version of paratest that is compatible to date, they will lose improvements resolutions and bugs but they will be able to use the old version anyway.

It is a security for systems using any software upgrade their tools, mainly to avoid the version hell.

Webysther:
@julianseeger I've been following the php-code-coverage of the improvements and they seem to better align namespace to correct problems that are essentially incompatible well.

I have noticed in recent weeks hundreds of tools holding the version that does not break, but losing the optimized handling of phpcov, which directly affects quality metric tools.

Webysther:
@michaelbutler Yes, is very important doing this.

julianseeger:
Ok that sounds like paratest `0.13.3` is for now and all times the one to use with the "old stable" phpunit 4.8 and the next release (either `0.14.0` or `1.0.0`) is the one to use with the "current stable" phpunit 5.4 ( according to the current https://phpunit.de ).

As far as I'm informed by @brianium it is desirable to get that working xml output pre `1.0.0`. So if anyone can confirm #19 fixed by #184 and there would be a fix for #199 (maybe just dropping that "line") we would be free to jump to 1.0.0. Isn't that a good incentive to fix #199? ;)

michaelbutler:
If I have time, I'll try to work on it this week :)

munkie:
Hey guys, any progress on this ?

aik099:
Maybe for BC reasons we don't typehint in methods for `\SebastianBergmann\CodeCoverage\CodeCoverage` instead of `\ PHP_CodeCoverage` but:

* remove typehint
* do a type check inside the method to allow either of classes

I want preserve BC, which PHP_CodeCoverage package apparently failed to do and that caused such problem.

munkie:
I have created PR that adds support for both phpcov versions 3rd and 4th #214

julianseeger:
Closing this in favor of https://github.com/brianium/paratest/pull/214
thanks @Webysther

PR #208, opened 2016-06-07
Fix failed build (Mb/MB check and require php-code-coverage 3.X)

michaelbutler:
# MB vs Mb string check

This should allow for differences in environments/versions which check the memory usage output string. Sometimes its X.XXMB and other times it's X.XXMb.

# PHP_CodeCoverage

Paratest depends directly on PHP_CodeCoverage via its code, however it does not explicity require it in composer.json. Instead it relies on the inclusion through phpunit. A recent version of phpunit completely changed the class structure of PHP CodeCoverage, so for now I simply did `composer require "phpunit/php-code-coverage:3.*"` to fix this. This [automatically] locks the project to [a maximum of] PHPUnit 5.3.5 which is the last version that supported this.

### DIFF
--- a/composer.json
+++ b/composer.json
@@ -11,7 +11,7 @@
         "ext-simplexml": "*",
         "ext-pcre": "*",
         "composer/semver": "~1.2",
-        "phpunit/php-code-coverage": "3.*"
+        "phpunit/php-code-coverage": "<=3.3.9"
     },
     "type": "library",
     "description": "Parallel testing for PHP",
--- a/composer.json
+++ b/composer.json
@@ -10,7 +10,8 @@
         "ext-reflection": "*",
         "ext-simplexml": "*",
         "ext-pcre": "*",
-        "composer/semver": "~1.2"
+        "composer/semver": "~1.2",
+        "phpunit/php-code-coverage": "3.*"
     },
     "type": "library",
     "description": "Parallel testing for PHP",
--- a/test/unit/ParaTest/Runners/PHPUnit/ResultPrinterTest.php
+++ b/test/unit/ParaTest/Runners/PHPUnit/ResultPrinterTest.php
@@ -128,7 +128,7 @@ class ResultPrinterTest extends ResultTester
         $this->assertRegExp(
             "/\n\nTime: [0-9]+([.][0-9]{1,2})? " .
             "(minute|minutes|second|seconds|ms)?," .
-            " Memory:[\s][0-9]+([.][0-9]{1,2})?Mb\n\n/",
+            " Memory:[\s][0-9]+([.][0-9]{1,2})?M[Bb]\n\n/",
             $header
         );
     }


michaelbutler:
Aaaaand now the latest version of PHPUnit changed something with the class names and/or loading...

michaelbutler:
Yeah, so the latest version of PHP CodeCoverage now uses namespaces, so the class PHP_CodeCoverage truly does not exist anymore :( 

https://github.com/sebastianbergmann/php-code-coverage/blob/900370c81280cc0d942ffbc5912d80464eaee7e9/ChangeLog-4.0.md

julianseeger:
PHP5.5 is EOL in 1 month (http://php.net/supported-versions.php) so no reason to support it any more after that imho.
The MB and Mb stuff is really bad test design. Of course we have to deal with that as it's in the codebase a lot of times but I hope they don't change it that often. I mean ... if they decided that Bytes are better than Bits or they mean Mebibytes or whatever, they will *probably* stick with that.

PHP-Code-Coverage is more interesting though. If I remember correctly, it should only be used in one class and one test or at least not more than a handfull of classes so it should be manageable to check for class existance and decide which strategy to use.
So for the 1 month of php5.5 support we wouldn't even need a branch. Just that version leftover in packagist and a >=php5.6 in our composer.json. But PHP-Code-Coverage may last longer in peoples repositories so if it's manageable, we should do it. It's not that much code.

Overall I'm not a fan of branches for that purpose. If we start like that, we may end up with a branch for every minor and some more patch releases of phpunit and it's components and we don't have the manpower to maintain all branches. So only the master would get updates and thats as good as just increasing the minimum dependency in the composer.json.

So thanks for the contribution, it will get in as soon as the other PR with phpcc support is in, we have to deal with that imho. If there is some good reason/feature to stop supporting php5.5 now, go for it.

michaelbutler:
hmm.. ok in that case I think i'll try to make this work with the latest PHPUnit 5.4.x via a switch like you suggested. Mainly because I have selfish reasons that I need to continue using PHP 5.5 a little while longer 😄  until the move to 7.0!

michaelbutler:
Success! OK, all I had to do to fix the code coverage issue is to composer require <=3.3.9 (since it changed in 4.0.0). I think this makes sense since Paratest itself uses the PHP_CodeCoverage class directly, so we can't rely on phpunit including whatever version of code coverage it wants.

julianseeger:
Thanks for the contribution. I would hope that a composer update can manage that without the need to delete the comoser.lock manually.

Ftr.: I see this as an intermediate solution to clarify the working dependencies until we can support hopefully both versions of phpcc or until phpcc 4.0 is old enough to justify dropping 3.x support.

michaelbutler:
Unfortunately, composer is not supposed to automanage changes to composer.json due to the purpose of the composer.lock file; generally projects are expected to commit their lock files. In Paratest, `composer.lock` is ignored, leaving it up to the end user to manage. 

Basically, composer looks at the lock file **first** when doing `composer install`, to install or update libraries. That way if you have something like 3.* in your composer.json and install, you get 3.5.5, and that 3.5.5 becomes "locked" even if the package is updated to 3.6, due to the lock file. If you either delete lock file or run `composer update` you will get 3.6.

julianseeger:
> generally projects are expected to commit their lock files

To stick with that naming, thats projects vs libraries. Projects are using libraries and thus, paratest would be a library just like phpunit and symfony. You won't find composer.lock files in their repositories because composer would ignore them anyways when some project depends on those libraries.
For projects it makes totally sense to commit the composer.lock to allow for each developer and especially the ci-server / deployment to have exactly the same version and it's enough to support a single set of dependencies. We need to support a range of dependencies. We might tighten them with tighter constraints (like the <= 3.3.9) but we can't lock them to a single supported version. If everyone would do that, nobody could resolve a dependency tree.

Btw.: this fix made two other PRs build, good job :+1:

PR #210, opened 2016-06-09
Allow files with all tests skipped

sroussey:
Prevent files containing only skipped tests from causing the whole process to fail. This happens often with group exclusions, for example. This still checks that it is a validly formed xml file.

### DIFF
--- a/src/ParaTest/Logging/JUnit/Reader.php
+++ b/src/ParaTest/Logging/JUnit/Reader.php
@@ -81,7 +81,7 @@ class Reader extends MetaProvider
      */
     public function hasResults()
     {
-        return $this->xml->count() > 0;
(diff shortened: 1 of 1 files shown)


src/ParaTest/Logging/JUnit/Reader.php
```diff
@@ -81,7 +81,7 @@ public function getSuites()
      */
     public function hasResults()
     {
-        return $this->xml->count() > 0;
+        return count($this->xml->xpath('//testsuites') === 1;
```
whsiao: @sroussey  You are missing a parenthesis for count().

julianseeger:
Your fix produced a merge conflict. Please resolve that by merging or rebasing to the current master.

PR #214, opened 2016-06-27
Add support for php code coverage lib of both 3 and 4 versions

### LINKED ISSUE #211: Paratest does not support phpunit >= 5.4.x
phpunit 5.4.x depends on php-code-coverage >= 4 (https://github.com/sebastianbergmann/phpunit/blob/5.4/composer.json) but paratest depends on php-code-coverage >=3,<4 so composer complains about the unsatisfiable deps:
```
20:26:06 [411.1MB/31.72s] Resolving dependencies through SAT
20:26:21 [412.2MB/46.56s] Dependency resolution completed in 14.833 seconds
20:26:21 [412.3MB/46.56s] Your requirements could not be resolved to an installable set of packages.
20:26:21 [412.3MB/46.56s] 
20:26:21   Problem 1
20:26:21     - Conclusion: remove phpunit/php-code-coverage 4.0.0
20:26:21     - Installation request for brianium/paratest dev-master -> satisfiable by brianium/paratest[dev-master].
20:26:21     - phpunit/phpunit 5.4.4 requires phpunit/php-code-coverage ^4.0 -> satisfiable by phpunit/php-code-coverage[4.0.0].
20:26:21     - phpunit/phpunit 5.4.3 requires phpunit/php-code-coverage ^4.0 -> satisfiable by phpunit/php-code-coverage[4.0.0].
20:26:21     - phpunit/phpunit 5.4.2 requires phpunit/php-code-coverage ^4.0 -> satisfiable by phpunit/php-code-coverage[4.0.0].
20:26:21     - phpunit/phpunit 5.4.1 requires phpunit/php-code-coverage ^4.0 -> satisfiable by phpunit/php-code-coverage[4.0.0].
20:26:21     - phpunit/phpunit 5.4.0 requires phpunit/php-code-coverage ^4.0 -> satisfiable by phpunit/php-code-coverage[4.0.0].
20:26:21     - phpunit/phpunit 5.4.4 requires phpunit/php-code-coverage ^4.0 -> satisfiable by phpunit/php-code-coverage[4.0.0].
20:26:21     - Conclusion: don't install phpunit/php-code-coverage 4.0.0
20:26:21     - Installation request for phpunit/phpunit 5.4.* -> satisfiable by phpunit/phpunit[5.4.0, 5.4.1, 5.4.2, 5.4.3, 5.4.4].
20:26:21 
```

munkie:
This PR add support for PHP Code Coverage library 4th version without dropping support of 3rd
Fixes #211


src/ParaTest/Coverage/CoverageMerger.php
```diff
@@ -19,26 +24,70 @@ public function addCoverage(\PHP_CodeCoverage $coverage)
...
+    /**
+     * Get coverage report generator
+     *
+     * @return CoverageReporterInterface
+     */
+    public function getReporter()
+    {
+        if ($this->coverage instanceof \PHP_CodeCoverage) {
+            return new CoverageReporter3($this->coverage);
+        } else {
```
michaelbutler: (minor) simply remove the `else {` here and `}` after since the previous block returns if it ran.

src/ParaTest/Coverage/CoverageMerger.php
```diff
@@ -2,14 +2,19 @@
 
 namespace ParaTest\Coverage;
 
+use SebastianBergmann\CodeCoverage\CodeCoverage;
```
michaelbutler: interesting... so this does not cause an error if you don't have the latest CodeCoverage installed?

src/ParaTest/Runners/PHPUnit/BaseRunner.php
```diff
@@ -1,7 +1,9 @@
 <?php
+
 namespace ParaTest\Runners\PHPUnit;
 
 use ParaTest\Coverage\CoverageMerger;
+use ParaTest\Coverage\CoverageReporter3;
```
michaelbutler: looks like this can safely be deleted, i don't see it anywhere in this file, PHPDoc or otherwise.

michaelbutler:
👍 overall this looks great! Thank you for implementing it.

src/ParaTest/Coverage/CoverageReporter4.php
```diff
@@ -0,0 +1,56 @@
+<?php
+
+namespace ParaTest\Coverage;
+
+use SebastianBergmann\CodeCoverage\CodeCoverage;
+use SebastianBergmann\CodeCoverage\Report\Clover;
+use SebastianBergmann\CodeCoverage\Report\Html\Facade;
+use SebastianBergmann\CodeCoverage\Report\PHP;
+
+class CoverageReporter4 implements CoverageReporterInterface
```
michaelbutler: minor... Maybe name this just `CoverageReporter` since it is the latest and greatest for the foreseeable future, but keep `CoverageReporter3` as is since it is "legacy"

src/ParaTest/Coverage/CoverageMerger.php
munkie: Nope. No problem if class in use is missing.

munkie:
@michaelbutler Pushed code review changes.

michaelbutler:
Nice job on this. let's see what @julianseeger has to say!

julianseeger:
Looks good to me. It even makes the BaseRunner a lot cleaner by moving the coverage specific code.
How does `testSimpleMerge` behave with phpcc3 (and testSimpleMergeLegacy on phpcc4)?

This seems to match the [suggestion](https://github.com/brianium/paratest/pull/207#issuecomment-228770855) of @aik099 . What about the concerns of @michaelbutler that it's
> kind of crappy reducing to a loose interface without typehints

and of @Webysther about
> but losing the optimized handling of phpcov, which directly affects quality metric tools.

?
The loose interface is true. Imho not that bad because paratest is a wrapper, nothing that much code will integrate. And @michaelbutler , you have dropped that opinion?

And about the quality metrics: [scrutinizer](https://scrutinizer-ci.com/g/brianium/paratest/inspections/38b6e252-bc2a-44cb-a981-fa08342c8422) is pretty much ok with it. One "major issue" because it can't find both php coverage classes at the same time but I'm ok with hiding that issue ^^. Is that what you meant @Webysther ?

munkie:
@julianseeger @michaelbutler I've added functional tests for CoverageMerger and CoverageReporter's.

src/ParaTest/Coverage/CoverageMerger.php
```diff
@@ -2,43 +2,80 @@
...
+     * Returns coverage object from file.
+     *
+     * @param \SplFileObject $coverageFile Coverage file.
+     *
+     * @return \PHP_CodeCoverage|CodeCoverage
      */
-    private function setCoverage(\PHP_CodeCoverage $coverage)
+    private function getCoverageObject(\SplFileObject $coverageFile)
     {
-        $this->coverage = unserialize(serialize($coverage));
```
munkie: I've removed this `unserialize(serialize())` trick. Did not find any comment why it is used and since tests are passing with different php and phpunit versions decided to remove it cause it can reduce performance.

julianseeger: maybe that's the reason why the hhvm tests are failing now

src/ParaTest/Coverage/CoverageMerger.php
```diff
@@ -2,43 +2,80 @@
...
+        if ('<?php' === $coverageFile->fread(5)) {
+            return include $coverageFile->getRealPath();
```
munkie: Instead of loading whole coverage file content, only first line is checked for php tag - should reduce memory usage  and improve performance.

michaelbutler:
I think loosening the typehints is not the end of the world here so I'm OK
with it.  I think the only way around that besides how it is done now is to
implement wrapper classes around the new CodeCoverage and the old
PHP_CodeCoverage,  then have them implement an interface, then typehint on
an interface. That might be more trouble than it is worth though!

src/ParaTest/Coverage/CoverageMerger.php
munkie: No HHVM are ok, they were failing because test did not have skip if code coverage not enabled.

munkie:
@michaelbutler So is it OK to be merged ?

michaelbutler:
Looks good to me but I don't have merge powers here.

julianseeger:
Personally, I don't like using real dumps like the `.cov` files because they hide the intent. You never know what part of the content matters unless you look at every test using it and in this special case the content doesn't even matter. So one could strip it down to a handful of lines. But just my flavor I guess ;)

Thanks for the contribution, I'm glad that we can stay compatible and the code looks even better than before imho.

munkie:
Cool!

PR #217, opened 2016-07-19
Support PHPUnit its --repeat option

PieterDC:
Provide ParaTest support for PHPUnit its repeat option.

We use the repeat option mostly to run tests on fixes for random failing tests. To prove that the tests succeeds X times in a row. Those are mostly single tests. Then we don't need ParaTest. PHPUnit suffices. But this pull request provides some groundwork for the upcoming support for --only-repeat-failed in ParaTest. An option we already contributed to PHPUnit. See https://github.com/sebastianbergmann/phpunit/pull/1965

### DIFF
--- a/src/ParaTest/Runners/PHPUnit/BaseRunner.php
+++ b/src/ParaTest/Runners/PHPUnit/BaseRunner.php
@@ -59,6 +59,13 @@ abstract class BaseRunner
      */
     protected $coverage = null;
 
+    /**
+     * Indicates whether the starting messages was printed or not.
+     *
+     * @var bool
+     */
+    protected $starting_messages_printed = false;
+
 
     public function __construct($opts = array())
     {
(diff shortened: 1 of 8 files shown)


src/ParaTest/Runners/PHPUnit/BaseRunner.php
```diff
@@ -59,6 +59,13 @@
      */
     protected $coverage = null;
 
+    /**
+     * Indicates whether the starting messages were printed or not.
+     *
+     * @var bool
+     */
+    protected $starting_messages_printed = false;
```
julianseeger: please stick to lower camel case ($startingMessagePrinted)

src/ParaTest/Runners/PHPUnit/BaseRunner.php
```diff
@@ -72,7 +79,12 @@ public function run()
         $this->verifyConfiguration();
         $this->initCoverage();
         $this->load();
-        $this->printer->start($this->options);
+
+        // Print the starting messages only if they were not printed yet.
+        if (!$this->starting_messages_printed) {
```
julianseeger: you don't need that comment if you just extract that code to a new method. The next dev would then read
`$this->printStartingMessageIfNotPrintedYet();`
... that's at least stable for changes and refactorings

julianseeger:
Thanks for the contribution.
Where are the tests for this? Please link or add some

PR #218, opened 2016-07-19
Support PHPUnit its --only-repeat-failed option

PieterDC:
PHPUnit doesn't have an option to only repeat failed tests yet. But we provided a pull request for it: https://github.com/sebastianbergmann/phpunit/pull/1965

And we coded ParaTest support for --only-repeat-failed, building upon what has been done for the --repeat option: https://github.com/brianium/paratest/pull/217


julianseeger:
Wow what happened to your history? That's 8 lines per commit. I guess that's what happens when you really apply commit early, commit often ;)

For the phpunit PR: I think you need to direct every PR to the master. It then get's merged to all maintained releases (if they whish so).

julianseeger:
@PieterDC can you please give me an update on this? Your PR for phpunit has been closed and you didn't link any new one that targeted a current phpunit version. If that's no longer an option, I would close this and https://github.com/brianium/paratest/pull/217 ... even though a repeat-option would be nice.

julianseeger:
Without the --only-repeat-failed option in phpunit, there is nothing to support in paratest

PieterDC:
I haven't made progress on https://github.com/sebastianbergmann/phpunit/pull/1965 Also not in a separate issue. Also not on my local machine. All code I have/had concerning that issue is pushed.

I agree that
> Without the --only-repeat-failed option in phpunit, there is nothing to support in paratest

So the PHPUnit feature request ideally has to be finished first.

PR #223, opened 2016-10-28
Fix issue 199, prevent empty line attributes in XML

michaelbutler:
# What

Fixes issue https://github.com/brianium/paratest/issues/199, where the paratest xml writer was allowing empty line numbers, e.g. `line=""`. This is probably due to PHP's loose typing, an empty value could be written as blank instead of `0` which is what we want.

I've written a unit test that uses a fixture xml with some `line=""` instances, and confirms that after writing the output, there is no `line=""` and there is at least one `line="0"` to confirm it changed properly.

The actual code fix is simply to check for an empty string at constructor time, and sets it to `0` if so.
## Side note:

This was created at Digital Ocean's Hacktoberfest in NYC.  (https://hacktoberfest.digitalocean.com/)

### DIFF
--- a/src/ParaTest/Logging/JUnit/Writer.php
+++ b/src/ParaTest/Logging/JUnit/Writer.php
@@ -133,6 +133,10 @@ class Writer
         $vars = get_object_vars($case);
         foreach ($vars as $name => $value) {
             if (preg_match(static::$caseAttrs, $name)) {
+                if ($name === 'line' && empty($value)) {
+                    // Don't output blank line attributes
+                    continue;
+                }
                 $caseNode->setAttribute($name, $value);
             }
         }
--- a/src/ParaTest/Logging/JUnit/Writer.php
+++ b/src/ParaTest/Logging/JUnit/Writer.php
@@ -133,8 +133,7 @@ class Writer
         $vars = get_object_vars($case);
         foreach ($vars as $name => $value) {
             if (preg_match(static::$caseAttrs, $name)) {
-                if ($name === 'line' && empty($value)) {
-                    // Don't output blank line attributes
+                if ($this->isEmptyLineAttribute($name, $value)) {
                     continue;
                 }
                 $caseNode->setAttribute($name, $value);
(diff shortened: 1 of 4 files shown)


michaelbutler:
@julianseeger Does Scrutinizer still run on this repo? It could help keep the code style consistent.

mabu-github:
As the line attribute is optional, you might want it not to get rendered out at all. 0 would indicate that a line number could be determined which was not the case.

michaelbutler:
That may be a good idea. I've done that now by adding a check right before the XML attribute is appended.

julianseeger:
https://scrutinizer-ci.com/g/brianium/paratest/inspections/5d2bda90-85d1-4620-b7ee-a2c90ed6d589
> No notable changes
There were no notable changes in the changeset which was inspected.

so it's either not working anymore or it has nothing to say ;)

src/ParaTest/Logging/JUnit/Writer.php
```diff
@@ -133,6 +133,10 @@ protected function appendCase($suiteNode, TestCase $case)
         $vars = get_object_vars($case);
         foreach ($vars as $name => $value) {
             if (preg_match(static::$caseAttrs, $name)) {
+                if ($name === 'line' && empty($value)) {
+                    // Don't output blank line attributes
+                    continue;
+                }
```
michaelbutler: Oh, great point.

julianseeger:
>This was created at Digital Ocean's Hacktoberfest in NYC. (https://hacktoberfest.digitalocean.com/)

good luck then ;)

I'm fine with these changes, thanks for that. I would prefer the small refactoring I mentioned.
Do you want to wait for feedback from #199 or shall I just merge?

michaelbutler:
I'll do the small refactor in this same PR. Let's wait a bit to see if the user responds.

michaelbutler:
OK this should be good to go now. I made the small change. At this point we probably don't need to wait for the orig. user's feedback. 🎃 👻

julianseeger:
For NYC, it should still be october, shouldn't it? ;)

michaelbutler:
Haha, yes! No worries though I didn't open enough prs 😂

PR #224, opened 2016-12-23
Fix temp dir inconsistency on os x

elliotfehr:
On many versions of OS X (including 10.12.1) there are inconsistencies between the temp directory returned by `sys_get_temp_dir()` and the one used by `tempnam()`.

`sys_get_temp_dir()` will return something like:
`/var/folders/0s/rlfsn9993r78qq7qjkr42x2w00058k/T`

`tempname()` will prefix with `/private` as the return value of `sys_get_temp_dir()` may not be writable:
`/private/var/folders/0s/rlfsn9993r78qq7qjkr42x2w00058k/T/paratest_oiqMU3`

This updates the test to take the directory that is returned by `tempnam()`, allowing tests to pass on OSX.

### DIFF
--- a/test/functional/PHPUnitTest.php
+++ b/test/functional/PHPUnitTest.php
@@ -350,7 +350,8 @@ class PHPUnitTest extends FunctionalTestBase
         if ($coveragePhp) {
             $this->assertEquals($coveragePhp, $options['coverage-php']);
         } else {
-            $this->assertStringStartsWith(sys_get_temp_dir() . '/paratest_', $options['coverage-php']);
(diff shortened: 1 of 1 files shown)


julianseeger:
We don't support OSX.

Just kidding ^^ Looks good to me. Thanks fot he contribution. The failing test is another one (I re-run the master and it seems to be failing, too) so not the concern of this PR but the master needs to be fixed first.

julianseeger:
Thanks for the contribution.
If you know how to work with travis and osx, feel free to extend the build matrix (.travis.yml) to keep paratest running on osx in the future.

PR #225, opened 2016-12-28
Allows groups to be specified, even if test suites do not contain tests for them

### LINKED ISSUE #164: Error with group option

t11n:
This fixes #164 and #146 

There are several bugfix proposals, none of them has been accepted so far. This approach is different and allows empty result files to be accepted, no matter what. Seems to be fine, according to tests. In case an empty result is processed, zero and empty values are accumulated, so results should not be affected. Overall count of suites is increased, but this is true, even if no tests are executed from these suites.

Please feel free to leave comments.

### DIFF
--- a/src/ParaTest/Runners/PHPUnit/ResultPrinter.php
+++ b/src/ParaTest/Runners/PHPUnit/ResultPrinter.php
@@ -201,16 +201,6 @@ class ResultPrinter
                 $test->getPath()
             ));
         }
-        if (!$reader->hasResults()) {
-            throw new \RuntimeException(sprintf(
-                "The process: %s\nLog file \"%s\" is empty.\n" .
-                "This means a PHPUnit process was unable to run \"%s\"\n" .
-                "Maybe there is more than one class in this file.",
-                $test->getLastCommand(),
-                $test->getTempFile(),
-                $test->getPath()
-            ));
-        }
         $this->results->addReader($reader);
         $this->processReaderFeedback($reader, $test->getTestCount());
         $this->printTestWarnings($test);
--- a/src/ParaTest/Logging/JUnit/Reader.php
+++ b/src/ParaTest/Logging/JUnit/Reader.php
@@ -216,6 +216,8 @@ class Reader extends MetaProvider
 
         if ($node !== false) {
             $this->suites[] = TestSuite::suiteFromNode($node);
+        } else {
+            $this->suites[] = TestSuite::suiteFromArray(self::$defaultSuite);
         }
     }
 }
--- a/src/ParaTest/Logging/JUnit/Reader.php
+++ b/src/ParaTest/Logging/JUnit/Reader.php
@@ -74,17 +74,6 @@ class Reader extends MetaProvider
     }
 
     /**
-     * Checks whether the xml log has
-     * test suite results
-     *
-     * @return array
-     */
-    public function hasResults()
-    {
-        return $this->xml->count() > 0;
-    }
-
-    /**
      * Return an array that contains
      * each suite's instant feedback. Since
      * logs do not contain skipped or incomplete
--- a/src/ParaTest/Parser/Parser.php
+++ b/src/ParaTest/Parser/Parser.php
@@ -119,7 +119,8 @@ class Parser
             }
         }
 
-        // Test class was loaded before somehow (referenced from other test class, or explicitly loaded)
+        // Test class was loaded before somehow
+        // (referenced from other test class, explicitly loaded, or filename does not match classname)
         foreach ($classes as $className) {
             $class = new \ReflectionClass($className);
             if ($class->getFileName() == $filename) {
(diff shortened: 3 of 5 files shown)


src/ParaTest/Runners/PHPUnit/ResultPrinter.php
```diff
@@ -201,16 +201,6 @@ public function printFeedback(ExecutableTest $test)
                 $test->getPath()
             ));
         }
-        if (!$reader->hasResults()) {
```
michaelbutler: I don't think we can safely remove this condition since it's designed to detect certain edge cases (like a process crashed, or there is wrongly more than one class in the file, etc...)

t11n: Thanks for the feedback. I looked into it. The reader still detects process crashes. It checks if the file size is 0 in ParaTest\Logging\JUnit\Reader on line 46, whereas hasResult() checks whether there is empty XML (but valid) in the file. If it is invalid XML, instantiating the SimpleXmlElement object would fail (which is not caught by the way, so there is probably no use case for that).

An empty XML file indicates there was no test executed. An entirely empty file or invalid XML lets one assume the process crashed or went mad or something.

I do not really understand what you mean with "more than one class in the file". Do mean several test cases in one unit test file? I tried both, having several unit tests in one test file, which results in only the first being executed and I tried to place a second class in the file of the class to test. This did not effect results at all.

At least, this is what my quick tests with phpunit 5.7 yielded. It's likely I oversaw something.

Could you please point me to such edge cases?

It is really a pity, as we introduced test groups in our project to be able to run specific tests only to reduce overall execution time. But this is of no use, as we cannot use paratests anymore and have to rely on phpunit only instead. Less tests are executed, but they are not run simultaneously anymore. You probably get the issue here :-)

michaelbutler: Thanks a lot for your explanation. Of course I really want to get the groups thing working as well, so I'll try to think up some ways to "break" it in this branch and get back to you. 

On the other hand, instead of fixing this in the post-phase -- XML/log parsing -- I wonder if it could be fixed **up front** by simply excluding the entire test PHP file from even running if it is not in the requested group, or inversely has no tests NOT in the exclude group. 

Paratest actually does a PHP file parse before it even invokes any test, in order to get the test methods used in the file. The file is `\ParaTest\Parser\Parser`, which is used by the `initSuites` function in SuiteLoader.

michaelbutler:
Also, I've fixed the Travis CI build in a separate PR, so once that is approved & merged you can pull in the changes here and should get a passing build.

src/ParaTest/Runners/PHPUnit/ResultPrinter.php
t11n: Probably one could exclude test suites up front. I am a little bit worried to rebuild the logic of excluding tests from phpunit, though. We would have to imitate the behaviour and always keep it in sync with future versions of phpunit. This means work.

On the other hand I wonder, if throwing an exception when the result file has no results in it is actually a bug. From the theoretical point, I cannot imagine any situation where there is an empty XML file, other than the test suite did not execute any tests (but finished cleanly).

In ParaTest\Logging\JUnit\Reader it is made sure the file is not empty, that it exists as well as that it holds well formed XML (very likely to be valid junit log XML). When constructor has passed, all these things are guaranteed.

Consequently, the ::hasResults() method only returns whether there are results from the test suite run or not. It does not say something about faulty states. It holds XML produced by phpunit, so phpunit always was able to write the junit log file.

Is it possible that throwing an exception at this point is some historical relict, that may have made sense in the past, but does not do anything now, but preventing us from using the group option?

The patch would help any issues related to test suites, that do not execute any tests. It is not strictly tight to the group option. Imagine a test suite that has tests for certain platforms. If you execute it on a different environment, the suite may not return any results. No problem for phpunit, but paratest would currently fail with this.

What do you think? Shell we try and treat the prior behaviour as faulty?

t11n:
Cool, I am happy to merge changes.

julianseeger:
Build is re-running, looks good so far.
Do I get your comments right that `$reader->hasResults()` is no longer required because it only handled cases that are no longer relevant with the xml/travis fix?

t11n:
Actually yes, the method is not used anymore. Removed it.

I did a little bit of research, too and found that the exception, as well as the hasResult method were introduced with this PR: https://github.com/brianium/paratest/pull/133 and refined in https://github.com/brianium/paratest/pull/135

It looks to me the exception should warn the developer if any suite is accidentally sorted out during parsing, if the name of the suite does not match its filename. However there is a fall back implemented in Parser::getClassName, so from the theoretical point it should not sort out anything, but really just prefer test suites whose name match their filenames. I added a note in the comments for that.

What do you guys think? Save to remove that exception?

PR #228, opened 2017-02-21
Fix Failing Build (Expected Exit Code)

michaelbutler:
# What

**1.** Trying to get this repo back in line. I think it's time to require phpunit 5.0+ (while not yet using PHPUnit 6.0 due to the namespace changes). I think this means dropping support for 5.3 as well. Composer.json has been updated.

**2.** The [testRunWithFatalParseErrorsHasExitCode255](https://travis-ci.org/brianium/paratest/jobs/172151038) test was failing on PHP 7.0 due to some change in either PHP itself, or some dependency that has changed. What it comes down to I think is that PHP 7.0 changed some fatals to exceptions, and symfony console made some changes accordingly to reflect this (?). `1` is still a "bad" exit code so it should be fine to check for either 255 or 1... but open to hearing opinions.

### DIFF
--- a/composer.json
+++ b/composer.json
@@ -2,7 +2,7 @@
     "name": "brianium/paratest",
     "require": {
         "php": ">=5.5.11",
-        "phpunit/phpunit": ">=3.7.8",
+        "phpunit/phpunit": "~5.0",
         "phpunit/php-timer": ">=1.0.4",
         "symfony/console": "~2.3|~3.0",
         "symfony/process": "~2.3|~3.0",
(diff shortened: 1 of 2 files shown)


michaelbutler:
Waiting on the builds to finish, to see what fails due to the PHPUnit change.

michaelbutler:
Oh, I forgot that paratest already requires version PHP 5.5 and up (and old version of PHPUnit "unofficially" work). I open this up for discussion and possible merging.

julianseeger:
Ah nice, thanks for this PR. Dropping support for phpunit < 5 meets phpunits own official support so thats fine. Maybe we should soon update the .travis.yml (support php 5.6 - 7.1).

Fixing phpunit 6.0 support is up to https://github.com/brianium/paratest/issues/226 now.

PR #233, opened 2017-04-13
Add gitattributes file

jdufresne:
Used by composer to ignore tests and other development files in the
distribution of a release.

### DIFF
--- a/.gitattributes
+++ b/.gitattributes
@@ -1,0 +1,8 @@
+/.gitattributes export-ignore
+/.gitignore export-ignore
+/.scrutinizer.yml export-ignore
+/.travis.yml export-ignore
+/Guardfile export-ignore
+/TODO.txt export-ignore
+/phpunit.xml.dist export-ignore
+/test/ export-ignore


julianseeger:
How do you guarantee that paratest runs without the test directory? Besides the test directory I don't see a real benefit in excluding those files from an export. The phpunit.xml.dist might even be quite useful to be delivered.

phpunit for example only excludes the build directory (over 6MB size) but still keeps the src (~1MB) and test (~2MB) dirs and all those small files.

Can you explain a bit more please why you choose these files and how one can check that paratest still works?

jdufresne:
> How do you guarantee that paratest runs without the test directory?

Not sure what you mean. Are you saying paratest can't be used to test _my project_ without its own tests in my `vendor` directory?

> The phpunit.xml.dist might even be quite useful to be delivered.

How so? When I install paratest through compser, I want to _use_ paratest, not develop paratest. For using paratest, its own `phpunit.xml` is of no use to me.

> phpunit for example only excludes the build directory (over 6MB size) but still keeps the src (~1MB) and test (~2MB) dirs and all those small files.

Planning to tackle phpunit at some point, but it has so many dependencies I haven't finished yet.

> Can you explain a bit more please why you choose these files and how one can check that paratest still works?

I chose to exclude files that are used for _developing_ paratest and choose to keep files that are for using (running) paratest. For example, any file I download into my `vendor` directory should be useful to me. Otherwise it is just a waste of bandwidth and bloat. When I want to develop paratest, I should clone the repo, not install through composer.

This blog post might help: https://blog.madewithlove.be/post/gitattributes/

Does this make sense?

julianseeger:
> Not sure what you mean. Are you saying paratest can't be used to test my project without its own tests in my vendor directory?

I hope not but you actually can't be sure. I mean ... I don't know it for sure, do you?
Someone could mess up the autoloader or add some require or include anywhere and you wouldn't notice because the tests are still green but maybe some class would crash if it's getting loaded without that test file it (badly) depends on.
Maybe that's a bit too theoretical here but I've actually seen that many times in other projects. Fortunately, paratest itself is not used in production systems but to test other code so it would 'only' crash some builds but not live systems. Without that limitation I - personally - wouldn't add such a gitattributes file to a project without at least some smoke tests that check if the code basically still works after an export.

That would be an interesting chat with a boss: 'So... why did that thing crash yesterday and made us loose 100k$?' 'uh well... we just wanted to save 70kb o.O'
Just to make the position clear ^^

But anyways, thanks for the contribution :)

jdufresne:
> I hope not but you actually can't be sure. I mean ... I don't know it for sure, do you?

I was curious about this as well. Just now, I ran a test in my project where after installing paratest, I did `rm -rf vendor/brianium/paratest/test` then reran my tests using paratest. It worked successfully.

Thanks!

PR #237, opened 2017-06-07
Create a release compatible with phpunit 6.x

### LINKED ISSUE #226: Create a release compatible with phpunit 6.0+
https://github.com/sebastianbergmann/phpunit/wiki/Release-Announcement-for-PHPUnit-6.0.0#backwards-compatibility-issues

They changed from underscores to namespaces for naming classes. I haven't checked if there are other incompatibilities.

TysonAndre:
I'd suggest bumping the major version to prevent accidental upgrades: The xml format is slightly different, phpunit removed deprecated functionality (if phpunit were to be installed as an indirect dependency), classes are different and so on.

- Could theoretically use `class_alias` depending on php version, but don't see a need to do so. Again, the xml format this outputs would be slightly different
- Not sure if I fixed `skipped` in all of the places
- changed the default behavior
- Removed old php and phpunit versions from travis.yml. Thoughts about adding phpunit 6.0 and 6.x separately?
- Updated example XML (didn't matter to unit tests. There should be no leading `\` for `PHPUnit\`... in xml)

Resolves #226

### DIFF
--- a/composer.json
+++ b/composer.json
@@ -2,7 +2,7 @@
     "name": "brianium/paratest",
     "require": {
         "php": ">=7.0",
-        "phpunit/phpunit": "~6.0.0",
+        "phpunit/phpunit": "~6.0",
         "phpunit/php-timer": ">=1.0.4",
         "symfony/console": "~2.3|~3.0",
         "symfony/process": "~2.3|~3.0",
--- a/composer.json
+++ b/composer.json
@@ -1,8 +1,8 @@
 {
     "name": "brianium/paratest",
     "require": {
-        "php": ">=5.5.11",
-        "phpunit/phpunit": "~5.0",
+        "php": ">=7.0",
+        "phpunit/phpunit": "~6.0.0",
         "phpunit/php-timer": ">=1.0.4",
         "symfony/console": "~2.3|~3.0",
         "symfony/process": "~2.3|~3.0",
--- a/test/functional/TestGenerator.php
+++ b/test/functional/TestGenerator.php
@@ -26,7 +26,7 @@ class TestGenerator
 
     private function generateTestString($testName, $methods=1)
     {
-        $php = "<"."?php\n\nclass $testName extends PHPUnit_Framework_TestCase\n{\n";
+        $php = "<"."?php\n\nclass $testName extends PHPUnit\Framework\TestCase\n{\n";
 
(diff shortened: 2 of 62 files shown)


test/functional/TestGenerator.php
```diff
@@ -26,7 +26,7 @@ public function generate($tests=1, $methods=1)
 
     private function generateTestString($testName, $methods=1)
     {
-        $php = "<"."?php\n\nclass $testName extends PHPUnit_Framework_TestCase\n{\n";
+        $php = "<"."?php\n\nclass $testName extends PHPUnit\Framework\TestCase\n{\n";
```
matthewnessworthy: I think you need to escape your backslashes
```
$php = "<"."?php\n\nclass $testName extends PHPUnit\\Framework\\TestCase\n{\n";
```

composer.json
```diff
@@ -1,8 +1,8 @@
 {
     "name": "brianium/paratest",
     "require": {
-        "php": ">=5.5.11",
-        "phpunit/phpunit": "~5.0",
+        "php": ">=7.0",
+        "phpunit/phpunit": "~6.0.0",
```
matthewnessworthy:
```
"phpunit/phpunit": "~6.0",
```

TysonAndre:
Issue 6037 of facebook/hhvm means that we require hhvm 3.19(.1?) or newer in phpunit 6.x+ for now (There are still likely some edge cases). Also see issue 2581 of phpunit

https://docs.hhvm.com/hhvm/installation/linux#ubuntu-14.04-trusty-supported-through-hhvm-3.18 - no prebuilt packages exist

matthewnessworthy:
@TysonAndre you've picked up a conflict, please resolve

julianseeger:
Sorry for the delay.
I pulled in https://github.com/brianium/paratest/pull/241 first so that we have that in a phpunit5 release.

I trust you @TysonAndre and @matthewnessworthy that the code is fine and will pull this in as soon as the test are running again.

TysonAndre:
Rebased and resolved merge conflicts with  #241 just now.

#241 added PHPUnit_Framework_TestCase in a few more places, and my last commit changes those to the namespaced versions

julianseeger:
Thanks for the contribution

PR #241, opened 2017-06-07
Allow multiple classes in one phpunit test file.

TysonAndre:
PHPUnit can properly handle this case without crashing.
Paratest should as well.

Before, Paratest would assume that the first class in the for loop was a
TestCase instance, and sometimes fail.

In several files in a project I work on,
there are helper classes alongside the test file,
used only in that test file.

Update test cases to reflect the new behaviour, require them to be
subclasses of PHPUnit_Framework_TestClass

- Keep giving priority to classes following PSR-0 naming

### DIFF
--- a/src/ParaTest/Parser/Parser.php
+++ b/src/ParaTest/Parser/Parser.php
@@ -110,23 +110,39 @@ class Parser
         $classes = get_declared_classes();
         $newClasses = array_values(array_diff($classes, $previousDeclaredClasses));
 
-        foreach ($newClasses as $className) {
-            $class = new \ReflectionClass($className);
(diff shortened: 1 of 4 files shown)


Slamdunk:
This could not be supported anymore in the near future, see https://github.com/paratestphp/paratest/pull/422

TysonAndre:
> In several files in a project I work on,
> there are **helper classes alongside the test file,**
> used only in that test file.

The helper classes in that project aren't classes that extend TestCase. I'll have to see if paratest works anyways in PHPUnit 9 with that PR

> Multiple test case classes (classes that extend TestCase) that are declared in a single sourcecode file are no longer supported

PR #244, opened 2017-06-15
Convert all PHP files to use short array syntax

jdufresne:
As ParaTest now requires PHP>=7.0, use modern short array syntax.

Files converted using Short Array Syntax Converter:

https://github.com/thomasbachem/php-short-array-syntax-converter


julianseeger:
Thanks for the contribution ;)

PR #245, opened 2017-06-16
Convert ParaTest to use PSR-4 autoloading

jdufresne:
As the next version of ParaTest contain backwards incompatible changes, I'd humbly like to suggest ParaTest use this rare opportunity to move from PSR-0 autoloading to PSR-4.

From [PHP-FIG](http://www.php-fig.org/psr/psr-0/):

> Deprecated - As of 2014-10-21 PSR-0 has been marked as deprecated. PSR-4 is now recommended as an alternative.

From [composer](https://getcomposer.org/doc/04-schema.md#autoload):

> PSR-4 is the recommended way since it offers greater ease of use (no need to regenerate the autoloader when you add classes).

This changes alters composer and the directory structure to use composer's PSR-4 autoloading.

Also allows for a shallower directory structure.

The directory `test/functional` continues to use PSR-0. This directory contained many non-namespaced files so converting did not initially work without additional changes. If you'd like, I could include those changes as part of this work, but as tests are not a public API, I see this as less pressing. If it were to happen at a later date, it would have no backwards incompatibility concerns.


julianseeger:
> The directory test/functional continues to use PSR-0. This directory contained many non-namespaced files so converting did not initially work without additional changes. If you'd like, I could include those changes as part of this work, but as tests are not a public API, I see this as less pressing. If it were to happen at a later date, it would have no backwards incompatibility concerns.

Feel free to add another PR. Not using namespaces at all is the major issue of that folder then ^^
And thanks for the contribution

PR #247, opened 2017-06-18
Removed unbounded constraint

GrahamCampbell:


GrahamCampbell:
Note the use of `^` rather than `~`.

PR #248, opened 2017-06-20
fix misc minor issues noticed by Phan static analyzer

TysonAndre:
- Add missing properties that were intended to be read-only
  `empty($options->filter)` was called, so added a definition of the
  magic method __isset() as well. (`empty($x->a)` is equivalent to `!isset($x->a) || !$x->a`)
- Fix inconsistencies between phpdoc and actual code.
- Add scalar type hints in phpdoc/real code
- Be specific about types parsed from phpunit xml output.
- stream_set_blocking expects bool, not int (Would be noticeable in strict mode)

### DIFF
--- a/src/Logging/JUnit/TestCase.php
+++ b/src/Logging/JUnit/TestCase.php
@@ -28,7 +28,7 @@ class TestCase
     public $file;
 
     /**
-     * @var string
+     * @var int
      */
     public $line;
 
@@ -38,27 +38,39 @@ class TestCase
     public $assertions;
 
     /**
-     * @var string
+     * @var string|float (a stringified float, from phpunit XML output)
      */
     public $time;
 
     /**
-     * Number of failures in this test case
+     * List of failures in this test case
      *
      * @var array
      */
     public $failures = [];
 
     /**
-     * Number of errors in this test case
-     * TODO: Not a number?
+     * List of errors in this test case
      *
      * @var array
      */
     public $errors = [];
 
+    /**
+     * List of errors in this test case
+     *
+     * @var array
+     */
     public $skipped = [];
 
+    /**
+     * @param string $name
+     * @param string $class
+     * @param string $file
+     * @param int $line
+     * @param int $assertions
+     * @param string $time
+     */
     public function __construct(
         $name,
         $class,
@@ -151,8 +163,8 @@ class TestCase
             (string) $node['name'],
             (string) $node['class'],
             (string) $node['file'],
-            (string) $node['line'],
-            (string) $node['assertions'],
+            (int) $node['line'],
+            (int) $node['assertions'],
             (string) $node['time']
         );
 
--- a/src/Logging/JUnit/TestCase.php
+++ b/src/Logging/JUnit/TestCase.php
@@ -56,11 +56,7 @@ class TestCase
      */
     public $errors = [];
 
-    /**
-     * List of errors in this test case
-     *
-     * @var array
-     */
+    /** @var array */
     public $skipped = [];
 
     /**
--- a/src/Logging/JUnit/TestSuite.php
+++ b/src/Logging/JUnit/TestSuite.php
@@ -73,7 +73,7 @@ class TestSuite
      * @param int $failures
      * @param int $skipped
      * @param float $time
-     * @param ?string $file
+     * @param string|null $file
      */
     public function __construct(
         $name,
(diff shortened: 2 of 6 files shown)


src/Logging/JUnit/TestCase.php
```diff
@@ -38,27 +38,39 @@ class TestCase
...
-     * Number of errors in this test case
-     * TODO: Not a number?
+     * List of errors in this test case
      *
      * @var array
      */
     public $errors = [];
 
+    /**
+     * List of errors in this test case
```
julianseeger: ~~errors~~**skipped**

actually, these comments are amazingly obvious aren't they?
```
/** @var array */
public $skipped = [];
```
should do the job and even that annotation should be useless because the type is known by declaration

src/Logging/JUnit/TestSuite.php
```diff
@@ -66,6 +66,15 @@ class TestSuite
...
     public $cases = [];
 
+    /**
+     * @param string $name
+     * @param int $tests
+     * @param int $assertions
+     * @param int $failures
+     * @param int $skipped
+     * @param float $time
+     * @param ?string $file
```
julianseeger: I assume this means `@param string|null $file`.
Is there any spec for this? At least in the [phpdoc doc](https://phpdoc.org/docs/latest/guides/types.html) I can't find anything about it

src/Runners/PHPUnit/Options.php
```diff
@@ -115,6 +156,18 @@ public function __get($var)
     }
 
     /**
+     * Public read accessibility
+     * (e.g. to make empty($options->property) work as expected)
+     *
+     * @param string $var
+     * @return mixed
+     */
+    public function __isset($var)
```
julianseeger: If all properties are explicitly defined, is there still a need for magic methods?

src/Runners/PHPUnit/Options.php
```diff
@@ -183,7 +236,7 @@ protected static function vendorDir()
      * @param  array $options
      * @return array
      */
-    protected function filterOptions($options)
```
julianseeger: now this is triple-redundant.

The description (I don't think it makes sense) says it returns a collection, then we `@return array` and have the typehint.

Either 
 * a useful description-text + typehints in the signature or
 * annotated param and return doc with a description like `@return array of default options` + typehinted signature or
 * only the signature including typehints or
 * as it was before (worst option probably)

Or does that make any style checker crying ? ;)

src/Runners/PHPUnit/Options.php
TysonAndre: They're protected inside the class. __get and __isset are created to add **read-only** access outside the class. __set and __unset deliberately aren't

src/Runners/PHPUnit/Options.php
```diff
@@ -58,6 +58,47 @@ class Options
...
+    protected $noTestTokens;
+
+    /**
+     * @var bool
+     */
+    protected $colors;
+
+    /**
+     * Filters which tests to run.
+     * @var ?string
```
julianseeger: There are more of these nullable annotations. As you commented above, it's already on the phpdoc roadmap but I would suggest sticking with `string|null` until it's in. Probably this is only important for IDE based type detection as we don't have phpdoc in the ci IIRC

applies for every `\?[\w]+` in this diff ;)

src/Runners/PHPUnit/Options.php
julianseeger: It just looks inconsistent. I would either expect an ArrayAccess implementation with exceptions on set or real implemented getters but that might be personal preference and is definitely out of scope of this PR so nvm

julianseeger:
Remove the left `?` and this can get merged. Sorry for the delay
FTR.: the build is green (even though it's displayed as in progress here)

src/Runners/PHPUnit/Options.php
TysonAndre: Anyway, the code that made me notice this:

in this file:
```
 +    /**
 +     * @var string[]
 +     */
 +    protected $groups;
```

in SuiteLoader
```
236     private function testMatchGroupOptions($groups)
237     {
238         if (empty($groups)) {
239             return true;
240         }
241 
242         if (!empty($this->options->groups)
```

julianseeger:
Thanks for the contribution

PR #255, opened 2017-06-30
Rely on PHP_BINARY to get the right executable instead of ENV

Slamdunk:
Both `phpunit-wrapper` and `phpunit` rely on `#!/usr/bin/env php` hashbash to detect the php executable.
This can be not found in the `$PATH`, or a user can start paratest with a different php than the one of the path (for ex. with a command like `/my/custom/build/php vendor/bin/paratest`.

Relying on `PHP_BINARY` to execute both the script is the only solution.

I tried to make a test for this bug, but cannot find the way to disable Travis environment.

### DIFF
--- a/bin/phpunit-wrapper
+++ b/bin/phpunit-wrapper
@@ -1,4 +1,3 @@
-#!/usr/bin/env php
 <?php
 echo "Worker starting\n";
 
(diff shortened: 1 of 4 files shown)


julianseeger:
thanks for the contribution

Multiply:
Thank you so much for this. This has been so annoying on my Mac with php installed via brew, and some legacy php being installed at the same time.

PR #256, opened 2017-06-30
Add PHP-CS-Fixer rules

Slamdunk:
https://github.com/FriendsOfPHP/PHP-CS-Fixer is an amazing tools that automatically fix PHP code according to a rule set.
The rule set committed is a subset of the tool chose for itself: https://github.com/FriendsOfPHP/PHP-CS-Fixer/blob/master/.php_cs.dist

I've added the rule set check to travis build in order to force contributors apply the fixes too.


.travis.yml
```diff
@@ -24,12 +24,12 @@ env:
     - PHPUNIT_VERSION='*@dev'
 
 before_script:
-    - composer self-update
```
Slamdunk: This command is already done by Travis

.travis.yml
```diff
@@ -24,12 +24,12 @@ env:
...
     - composer require --no-update phpunit/phpunit=$PHPUNIT_VERSION
     - composer install --prefer-dist
 
 script:
     - vendor/bin/phpunit --coverage-clover=coverage.clover
+    - if [[ $TRAVIS_PHP_VERSION == "7.0" && $PHPUNIT_VERSION == '6.*@stable' ]]; then vendor/bin/php-cs-fixer fix --diff --dry-run --verbose ; fi
```
Slamdunk: Coding standard must be executed in the lower PHP version supported, but can be checked in one build only to speed up the build.
I suggest to move to a `- include` signature to have a simplier variable management; I can PR a PoC for this too.

Slamdunk:
The build **intentionally** fails in order to let the owner of this package inspect the changes exposed in the failing build:
https://travis-ci.org/brianium/paratest/jobs/248800471

julianseeger:
Ok this one is much more complicated than it looks like.
First of all - in advance to any critics - this is good. We have been all too lazy with code styles in this project and the scrutinizer integration wasn't as successful as I expected .. mabye of misconfiguration or misuse or because it's not available locally. So providing style checks including a fool-proof way to fix them is great.

### the easy part:

- fos/php-cs-fixer docs recommend creating a `.php_cs.dist` and reserve the gitignored `.php_cs` to the users/devs local configurations
- if multi-staged builds are available in public travis (according to the docs I guess so), that could be a way to make it more clear why the build failed
```
# stage "test":
 - ✔ 1
 - ✔ 2
 - ✔ 3

# stage "style":
 - X phpcs
```
(phpcs at the later stage because you know it's gonna succeed if you know that you care about it ^^)
- there should be an easy way for the users to reproduce that (i would suggest a composer script like `composer check` or `composer pre-commit`, don't know)

### then the not-so-easy part:

- I'm against ordering methods. While imports might be OK (assuming they are automatically managed by your favorite IDE most of the time), ordering class elements (especially methods) doesn't increase readability imho. In contrast, it can even decrease readability when private method that have been extracted by refactoring are moved away from the using public methods for example. Either you are jumping between methods with IDE support ... then you don't need any ordering at all ... or you aren't, then it is more useful when the methods are grouped by cohesion. And I don't feel like this is much personal preference but just a logical view to fixer induced readability damage o.O
- while it is easy to execute the fixer, the job it has to do should be minimized and thus there should be away to integrate detection tools like phpcs (which has wider support as far as a fast research tells me). While the php cs fixer doesn't support `phpcs.xml` configs (I'm a little bit disappointed by that), it seems possible to split the php-cs-fixer config into a php array of rules (like this one: https://github.com/squizlabs/PHP_CodeSniffer/blob/7e589ecc5b9cab2545ece03a8ddb84ab27c1824e/CodeSniffer.conf.dist ) and the `.php_cs.dist` using that config and returning a php-cs-fixer compatible instance. This way a single config / point of truth could support php-cs-fixer and phpcs (and thus a lot of other tools and integrations). Sounds reasonable?

### questions
- didn't find an explicit statement but the Symfony rule fulle includes all PSR2 rules doesn't it?
- why should anyone want legal notes in every file?
- `'phpdoc_add_missing_param_annotation' => true,`  what's the use of a param annotation without any explanation what that param is for? That sounds like those docblocks:
```
/**
 * The constructor
 * 
 * @param $opt
 */
public function __construct($opt) {
...
}

/**
 * Convert the object to a string
 */
public function __toString() {
...
}
```
Without any docblock you would loose nothing and gain readability. That works for all autogenerated doc blocks.

Slamdunk:
> fos/php-cs-fixer docs recommend creating a .php_cs.dist and reserve the gitignored .php_cs to the users/devs local configurations

Done

> if multi-staged builds are available in public travis (according to the docs I guess so), that could be a way to make it more clear why the build failed

Multi-staged builds are awesome, but are still a beta feature and I'm not much into them at the time of writing. I suggest to postpone this change in a future PR; if you agree you can create an issue as a reminder.

> there should be an easy way for the users to reproduce that (i would suggest a composer script like composer check or composer pre-commit, don't know)

If the step is well documented in README.md, I don't see a difference between `composer check` and `vendor/bin/php-cs-fixer fix`. To be honest, I prefer the latter because it is more clear that we use that particlar tool in its default behaviour; in the former case a curious contributor must open the `composer.json` file to get to know what we are doing.

> I'm against ordering methods. [...]

1. The rules written in this PR config are not rationally chosen yet. In my projects I have a very specific and aggressive [config](https://github.com/Slamdunk/php-cs-fixer-extensions/blob/v1.2.3/lib/Config.php) that do not fit a FOSS very well, so I accounted the rule that the tool itself use a good starting point, that I expected to be discussed and changed here.
1. `ordered_class_elements` is very helpful and can be deeply configured. Yes, reordering methods by visibility can be misleading, and can be disabled. But the primary target of this fixer is to order the *blocks* of code, for example all the traits, then all the constants, then all the properties, then all the methods. Having an order of the conceptual pieces of a class increase the readibility a lot.

> while it is easy to execute the fixer, the job it has to do should be minimized and thus there should be away to integrate detection tools like phpcs

I admit I don't understand what you mean, or better, I don't understand what you wrote 😸 
As a small recap, PHP_CodeSniffer is a tool born before PHP-CS-Fixer that, as the name suggest, was aimed to *detect* violations, and it evolved to *fix* them. PHP-CS-Fixer started immediately with the purpose to fix violations. Theoretically PHP_CodeSniffer can do more because it can tokenize JS and CSS, and you can write custom rules to detect edge cases that can't be fixed automatically; with PHP-CS-Fixer if it can't be fixed, it can't be detected.
That said, and said that the number of fixers in PHP-CS-Fixer is big and extremely configurable, I see no point in having a tool like PHP_CodeSniffer; github stars suggest that PHP-CS-Fixer is rapidly overtaking PHP_CodeSniffer, it's much easier to configure and faster to run.

> didn't find an explicit statement but the Symfony rule fulle includes all PSR2 rules doesn't it?

Yes: https://github.com/FriendsOfPHP/PHP-CS-Fixer/blob/v2.3.2/src/RuleSet.php#L55-L56

> why should anyone want legal notes in every file?

Because lawyers love quibbles, and FOSS maintainers don't like facing them.

> what's the use of a param annotation without any explanation what that param is for?

I have to say I don't use any docblock nor documentation in my project, because Tests are the only bible for me. But all FOSSs I know seem to love them, and it is supposed that, if a contributor creates a PR without docblock, the fixer and the reviewer should encourage him/her to write comments about them.

julianseeger:
> Multi-staged builds are awesome, but are still a beta feature and I'm not much into them at the time of writing. I suggest to postpone this change in a future PR; if you agree you can create an issue as a reminder.

#258

> If the step is well documented in README.md, I don't see a difference between composer check and vendor/bin/php-cs-fixer fix.

idc where, but one of the two choices (or both) should be in this PR. A Readme info that the fixer should be run in advance to ANY commit would be good anyways.

> The rules written in this PR config are not rationally chosen yet

Everything except for the `ordered_class_elements` sounds reasonable to me because it fits the style we already try to achieve. Besides that, I think a ruleset that creates a minimal diff is best for this PR so that additional rules can be added an discussed in separate PRs once this is merged and in use. I expect a lot of discussion in the next few PRs when contributors refuse to accept certain rules ^^ 

> ordered_class_elements is very helpful and can be deeply configured

Don't know how to configure it but something like:
`['use_trait', 'constant_public', 'constant_protected', 'constant_private', 'property_public', 'property_protected', 'property_private', 'construct', 'destruct', 'magic', 'phpunit', '*']`
should fit the project (derived from the defaults with * meaning "please don't care for the inner order of the remaining stuff). Some part of that is even included in PRS2. Is that doable?

> I admit I don't understand what you mean, or better, I don't understand what you wrote

Sorry for that ;) In short: however they started, the IDE integration for the php-cs-fixer is just bad - probably because the devs are sure their way to use it is the only one that matters. Because of that, it would be nice to include for example a phpcs config with all rules that exist in both tools. Using one tool doesn't mean you have to dump all others (Editorconfigs for example can be included in every project and are supported by every relevant IDE ...even `vim` ... even though they are not suffixient for every rule you want to define, it's a baseline that works out of the box).
**But that can be added "later".**

> Because lawyers love quibbles, and FOSS maintainers don't like facing them.

I'm no layer but as we don't have such messages in every file, I see no point in supporting such habits. If someone copies the code, he can also copy the code and remove that legal lines... but that's up to @brianium ( at least as long as  the repo is in his namespace :P )

> I have to say I don't use any docblock nor documentation in my project, because Tests are the only bible for me. But all FOSSs I know seem to love them, and it is supposed that, if a contributor creates a PR without docblock, the fixer and the reviewer should encourage him/her to write comments about them.

From my experience, most of the existing devs love comments. I'm supporting the uncle bob way of thinking that if you need inline comments, then your code is not readable enough and you should refactor it asap. If you need an descriptionary docblock, your method signature is bad (the method name is not discriptive enough and/or your parameters are not descriptive enough by themselves. Exceptions are if you generate API-Docs from these docblocks (especially with closed source) - what we don't do - and type annotations - that we (as I hope) don't need anymore because we have type hinting - and explanations for workarounds if you can't fix the broken code (because it's outside of the project) and it wouldn't be understandable without a comment.
But apart from that, the dry-run in travis looks good. Maybe we can get rid of all these docblocks some time in the future but thats going to be a religious discussion o.O

**things still to be adressed / discussed:**
* mention php-cs-fixer in the README
* don't order methods (or discuss why this is particularly useful ^^)
* do we need legal headers? @brianium

Slamdunk:
1. Command added to the README.md
1. Header removed
1. Methods fixer changed; now the only methods moved are a `__toString` and the phpunit `setUp`/`tearDown`.

If it's ok to you, I'm going to apply the fixes and squash the PR.

julianseeger:
looks good to me 👍

Slamdunk:
Fixers applied, PR ready to be merged.

julianseeger:
thanks for your work. and for taking the time for the discussion and fixups ;)

PR #259, opened 2017-07-04
Drop CodeCoverage < v5 support

Slamdunk:
Since Paratest v1 PHPUnit v6 is required (https://github.com/brianium/paratest/blob/1.0.0/composer.json#L4-L5), and since PHPUnit v6, CodeCoverage v5 is required (https://github.com/sebastianbergmann/phpunit/blob/6.0.0/composer.json#L26).

So we can drop any support on CodeCoverage < v5, infact those tests are always skipped.


julianseeger:
> Checked all files in 58.190 seconds, 10.000 MB memory used

10 GB? really? o.O

PR #260, opened 2017-07-04
Travis build refactored

Slamdunk:
Hi, I've refactored the Travis build with the following changes:

1. Dependencies are now installed on lowest or latest version; this is mandatory because if a version is allowed we need to test it
1. PHPUnit is now pinned to ^6.0.13 because there are some issues with intermediate minor versions, never tested before this PR and point 1 got emerged
1. Symfony is now pinned to ^3.0 for the same reason, Travis fails with Symfony ~2.0 but previous Travis config didn't show it
1. Code Coverage is done in one build only, on the lowest PHP Version but with the latest dependencies; an ENV variable is set to be clear on the build
1. CS Fixes are checked on an own build without tests involved; an ENV variable is set to be clear on the build
1. PHP 7.2 is now tested, but allowed to fail
1. PHPUnit@dev is tested on latest PHP only to reduce build matrix

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -2,35 +2,44 @@ language: php
 
 sudo: false
 
-dist: trusty
-
-# hhvm won't work with phpunit 6.x due to lack of Throwable support, see https://github.com/facebook/hhvm/issues/6037
-php:
-    - 7.0
-    - 7.1
-
 cache:
   directories:
-    - $HOME/.composer/cache
+    - $HOME/.composer/
 
 matrix:
   fast_finish: true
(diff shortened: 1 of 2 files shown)


Slamdunk:
travis-jobs

julianseeger:
Looks nice, thx.

> 1. Dependencies are now installed on lowest or latest version; this is mandatory because if a version is allowed we need to test it

This would actually mean we would have to test EACH version. Not only the highest and lowest. But I get the point... at least a way to find stuff like `2.`

PR #264, opened 2017-07-06
Add type hints and return types

Slamdunk:
Closes https://github.com/brianium/paratest/issues/263


julianseeger:
Thanks for the contribution.
Next step is removing `@param` and `@return` annotations from private methods' docblocks then?

PR #266, opened 2017-07-19
Suggest Composer CLI for package installation

mbrodala:


mbrodala:
Forgot a space :man_facepalming: see #271

PR #267, opened 2017-07-21
Update symfony dep to ^3.2, automatically inherit env variables

TysonAndre:
This may break assumptions in some projects about the way
paratest sets up environment variables.
(I think it might clear some of them, which differs from phpunit?)

Symfony 3.3 has deprecated the old behavior, which will be removed
eventually in 4.0.
(Saw this E_USER_DEPRECATED warning in a different project with a
different error_handler callback)

Symfony/console versions 3.0-3.1 don't implement inheritEnvironmentVariables

- To be consistent about environment variables, require a minimum of 3.2
- 3.0-3.1 are about to stop being supported
  https://symfony.com/roadmap?version=3.1#checker
- Keep the method_exists check for now, just in case 

See https://github.com/symfony/process/blob/3.3/Process.php#L319

### DIFF
--- a/composer.json
+++ b/composer.json
@@ -9,8 +9,8 @@
         "composer/semver": "~1.2",
         "phpunit/php-timer": "^1.0.4",
         "phpunit/phpunit": "^6.0.13",
-        "symfony/console": "^3.0",
-        "symfony/process": "^3.0"
+        "symfony/console": "^3.2",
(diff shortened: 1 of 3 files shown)


julianseeger:
No need to care about the last 10 days of Support for 3.1.
Thanks for the contribution

PR #270, opened 2017-07-23
Remove leading backslash from use statements

jdufresne:
Move conventional and inline with PHP recommendations.

From the PHP documentation:

http://php.net/manual/en/language.namespaces.importing.php#language.namespaces.importing

> Note that for namespaced names …, the leading backslash is unnecessary
> and not recommended, as import names must be fully qualified, and are
> not processed relative to the current namespace.

### DIFF
--- a/test/functional/FunctionalTestBase.php
+++ b/test/functional/FunctionalTestBase.php
@@ -2,7 +2,7 @@
 
 declare(strict_types=1);
 
-use \Symfony\Component\Process\Process;
+use Symfony\Component\Process\Process;
 
 class FunctionalTestBase extends PHPUnit\Framework\TestCase
 {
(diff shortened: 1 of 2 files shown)


julianseeger:
Can you add a higher version of phpcs-fixer https://github.com/brianium/paratest/blob/master/composer.json#L43 to your PR?
It's currently locked to 2.3.2 because 2.4.0 ist able to detect exactly your change.

jdufresne:
Doing so causes other phpcs errors in the files `src/Runners/PHPUnit/ExecutableTest.php` (braces) and `src/Runners/PHPUnit/Configuration.php` (no_break_comment) do you want these fixes included in this PR as well or would you prefer them to be separate?

julianseeger:
Sorry for the delay. Nvm, I just thought we would get it for free then, but didn't remember that it was more than the includes.

PR #272, opened 2017-07-28
Fix version reporting in console, closes #257

### LINKED ISSUE #257: The reported version in the console is wrong
When installing version 1.0.0 it reports back as 0.14.0.

Maybe this shouldn't be hardcoded in a file, but merely be read from the composer.json file, or something similar?

`php vendor/bin/paratest --version`

ktomk:
ParaTest has the version number hard-encoded.

Previously when reporting the version via --version that hard-encoded
version number was presented in the output (e.g. 1.0.1).

As technically the version changes with git commit and not every git
commit is tagged and not even every tagged commit contains the appropriate
change to the hard-encoded version number within

    \ParaTest\Console\ParaTestApplication::VERSION

the wrong version is reported.

This commit addresses the issue by first looking up the installed version
of the package "brianium/paratest" from within composer's installed.json
file which contains the currently installed version (based on the git tag,
see Packagist) and if not found (e.g. you have checked out the development
version in a git clone) the version number will be taken from git directly.

If everything of these look-ups fail, the fallback is to the current
behavior of taking the version from the constant.

Refs:

- #257

### DIFF
--- a/src/Console/ParaTestApplication.php
+++ b/src/Console/ParaTestApplication.php
@@ -19,7 +19,7 @@ class ParaTestApplication extends Application
 
     public function __construct()
     {
-        parent::__construct(static::NAME, static::VERSION);
+        parent::__construct(static::NAME, VersionProvider::getVersion(static::VERSION));
     }
 
     /**
--- a/src/Console/VersionProvider.php
+++ b/src/Console/VersionProvider.php
@@ -1,0 +1,107 @@
+<?php
+
+declare(strict_types=1);
+
+namespace ParaTest\Console;
+
+use Symfony\Component\Process\Process;
+
+/**
+ * Class VersionProvider.
+ *
+ * Obtain version information of the ParaTest application itself based on
+ * it's current installment (composer; git; default passed)
+ */
+final class VersionProvider
+{
+    const PACKAGE = 'brianium/paratest';
+    /**
+     * @var null
+     */
+    private $default;
+
+    public function __construct($default = null)
+    {
+        $this->default = $default;
+    }
+
+    public static function getVersion($default = null)
+    {
+        $provider = new self($default);
+
+        return $provider->getParaTestVersion();
+    }
+
+    public function getParaTestVersion()
+    {
+        return $this->getComposerInstalledVersion(self::PACKAGE)
+            ?? $this->getGitVersion()
+            ?? $this->default;
+    }
+
+    public function getGitVersion()
+    {
+        $version = null;
+
+        $process = new Process('git describe --tags --always --first-parent', __DIR__);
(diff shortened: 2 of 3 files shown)


ktomk:
With these changes, example output from development version (clone):
~~~
$ bin/paratest --version
ParaTest 1.0.1-10-g89f72e2
~~~
And when installed via composer:
~~~
$ vendor/bin/paratest --version
ParaTest dev-patch/give-version-257
~~~
It shows the inline alias version.

julianseeger:
Awesome 👍 
Should we change the hardcoded (default) version to 1.x then? That would make the output always correct until release 2.0.0 (correct even if not complete at the worst case like phar installs).

ktomk:
From what I know from other projects, it's OK to make it a place-holder like `@package_version@` (e.g. Composer does it that way IIRC).

When you start to provide a phar package, a build process can replace it.

Strings like the suggested "`1.x`" tend to be misleading in case things go havoc (too close to a "real" version when output so not obvious enough to the reader). It also has the benefit that these are easy to replace, so open up a good migration path.

For example, that special version as it is right now can be taken as-is for the moment, you pull in the pr and after some experience how it works out (it's hard to test, so usage will show up issues here you want to patch when found) - then after some time, you can consider to move to a place-holder-value there. I think that is a sane suggestion.

ktomk:
btw. and a bit off-record: How did you manage to merge the coverage? Just yesterday I was looking on how to invoke a PHP script within a test-suite. I was able to keep the parameters (in the end) for remote debugging, but code-coverage is quite a different beast as PHPUnit knows nothing about that child-process code-coverage.

julianseeger:
Thanks for the contribution.

The merging is a feature of php-code-coverage itself. I had to deliver a small fix that it works correct when you import an exported coverage but apart from that, it's as easy as $coverage1->merge($coverage2).
I think it just adds all lines from $coverage2 to $coverage1 and if a line already exists, a covered line is preferred.

Getting coverage of generic php subprocesses would be hard. But as paratest is executing multiple phpunit instances, we let each phpunit subprocess generate its own coverage and just need to collect these coverage exports...and merge them.

Multiply:
Bit late to the show, but relying on having `git` available, might not be the best thing to do.

For example, I install things using my own tools on OSX, or inside a docker container. When I actually run the code, it'll be in a separate container, where the `git`-executable is not available.

ktomk:
@Multiply: In case the `git` command fails, there is the fallback to the current behavior.

ktomk:
@julianseeger: Thanks for the infos. As my scenario is also about Phpunit, that kind of infomration is really helpful. Thanks.

PR #275, opened 2017-08-28
README.md: add code-coverage badge

Slamdunk:
I consider CC an info with higher priority of scrutinizer score.


julianseeger:
> I consider CC an info with higher priority of scrutinizer score.

Thats debatable because covered code doesn't imply that it's really tested (only executed during tests). But its a hint and doesn't hurt anyways.

Thanks for the contrib

Issue #158, opened 2015-02-25
Generic exceptions do not hint


tomaszdurka:
It's my first day with paratest and I already cannot really get it working with one of my projects.

All I get after running paratest is:
```
bin/paratest -c phpunit.xml

Running phpunit in 5 processes with phpunit

Configuration read from /home/vagrant/cm/phpunit.xml

  [InvalidArgumentException]                                                   
  Log file /tmp/PT_d13SdD is empty. This means a PHPUnit process has crashed.  
                                                                               
```
This is very generic error and doesn't tell anything about actual problem. There is no hint regarding why it crashes at all. To understand what is actually wrong you need to go into the code and/or debug.

I still am not sure what is actually wrong with this. Tests are running fine wihtout paratest. I think it would be great to actually catch some of those low level exceptions and rethrow some more descriptive one.

What can be wrong in this case?

tomaszdurka:
I've debugged a little bit and seems that runner starts 5 testcases, but none of them is actually creating junit log even though log-junit parameter is specified correctly for command. 

It's starting tests with following commands:
```
'bin/phpunit' '--configuration' '/home/vagrant/cm/phpunit.xml' '--log-junit' '/tmp/PT_x25ZsW' 'CM_Action_AbstractTest' '/home/vagrant/cm/tests/library/CM/Action/AbstractTest.php'
'bin/phpunit' '--configuration' '/home/vagrant/cm/phpunit.xml' '--log-junit' '/tmp/PT_yyThBm' 'CM_Action_ActionTest' '/home/vagrant/cm/tests/library/CM/Action/ActionTest.php'
'bin/phpunit' '--configuration' '/home/vagrant/cm/phpunit.xml' '--log-junit' '/tmp/PT_etI2JM' 'CM_Action_EmailTest' '/home/vagrant/cm/tests/library/CM/Action/EmailTest.php'
'bin/phpunit' '--configuration' '/home/vagrant/cm/phpunit.xml' '--log-junit' '/tmp/PT_mgMNTc' 'CM_AdproviderAdapter_AbstractTest' '/home/vagrant/cm/tests/library/CM/AdproviderAdapter/AbstractTest.php'
'bin/phpunit' '--configuration' '/home/vagrant/cm/phpunit.xml' '--log-junit' '/tmp/PT_I3N95C' 'CM_AdproviderAdapter_OpenxTest' '/home/vagrant/cm/tests/library/CM/AdproviderAdapter/OpenxTest.php'
```

[...]

PR #281, opened 2017-10-04
Suggest enabling xdebug for coverage

MarkVaughn:
When xdebug is disabled, the message why paratest fails is a bit cryptic.

This PR adds a check to see if xdebug is enabled and changes the message to be more specific if it's disabled/missing.

Related to #158


julianseeger:
thanks for the contribution, lgtm

PR #287, opened 2017-12-01
Add support for phpdbg, fixes #277

### LINKED ISSUE #277: phpdbg support
Currently it's not possible to run paratest with phpdbg instead of xdebug (dead slow) to generate coverage.

Running:
```bash
phpdbg -qrr vendor/bin/paratest tests
```

Results in wrong subprocess command construction as:
`/usr/bin/phpdbg7.1 /app/vendor/bin/phpunit --configuration=/app/phpunit.xml.dist --log-junit=/tmp/PT_I8omPT Tests\....`

It should be:
`/usr/bin/phpdbg7.1 -qrr /app/vendor/bin/phpunit --configuration=/app/phpunit.xml.dist --log-junit=/tmp/PT_I8omPT Tests\....`

I'm not familiar with this tool code, but looks like the problem lays out in [ExecutableTest](https://github.com/brianium/paratest/blob/master/src/Runners/PHPUnit/ExecutableTest.php#L191) by using PHP_BANARY constant.

It looks to me that it could be easily solved by replacing the constant with PhpExecutableFinder::find from symfony/process component.

mdevlamynck:


jhoff:
Confirmed that this works.

julianseeger:
Can you confirm that this works even with older versions of symfony (like 3.4)? 
This change: https://github.com/symfony/process/commit/067046d5f8209faea3b81196a1bb02c575d30d7f#diff-e314561824791ffd246d04bc2c3b4d9c 
indicates that there have reviously been issues that PHP_BINARY wasn't respected well (without using phpdbg)...

mdevlamynck:
@julianseeger We used it with symfony 2.8 and recently with 3.4 (because we use the LTS versions).

julianseeger:
ok, thanks for the contribution

PR #292, opened 2017-12-22
Make SuiteLoader more extendable

jondmcelroy:
I had an issue where I needed to overwrite the SuiteLoader for a new runner that I was created. Although it has some protected variables, most of the functions were private.

This PR is to update all of the functions to protected and to have the load function accept a loader.

### DIFF
--- a/src/Runners/PHPUnit/BaseRunner.php
+++ b/src/Runners/PHPUnit/BaseRunner.php
@@ -91,7 +91,6 @@ abstract class BaseRunner
      * to run. If functional mode is enabled $this->pending will
      * contain a collection of TestMethod objects instead of Suite
      * objects.
-     * @param SuiteLoader $loader
      */
     protected function load(SuiteLoader $loader)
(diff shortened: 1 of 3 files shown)


julianseeger:
About the build: I think phpcs is not ok with your phpdoc changes:
> 1) src/Runners/PHPUnit/BaseRunner.php (phpdoc_separation)

About the diff: did you ever think of just overwriting the public `load` function? Sounds more reasonable to me than overwriting ALL other methods of the SuiteLoader (API-Design wise).

jondmcelroy:
My main goal was to overwrite the Suite class. To accomplish this currently I am overwriting the TestRunner->load function so that I can create a new DifferentSuiteLoader which creates a my new DifferentSuite class. Since all of the functions are private in SuiteLoader, I have had to copy the entire class just to change the suite type that is being created, even if all I wanted to do was overwrite the SuiteLoader->load function.

PR #294, opened 2018-01-24
Fixed problem composer installs with bin-compat full

porebskk:
This will fix the method for determining the phpunit path, when installing e.g. on windows with the composer configuration "bin-compat": "full" and running tests inside a linux docker.  
The problem was that the current sollution determines the phpunit executable file based on the existence of the ".bat"-file. This solution will determine the executable based on file existence and the systems DIRECTORY_SEPERATOR.

### DIFF
--- a/src/Runners/PHPUnit/Options.php
+++ b/src/Runners/PHPUnit/Options.php
@@ -205,7 +205,7 @@ class Options
      */
     protected static function phpunit(): string
     {
-        $vendor  = static::vendorDir();
+        $vendor = static::vendorDir();
 
         $phpunit = $vendor . DIRECTORY_SEPARATOR . 'bin' . DIRECTORY_SEPARATOR . 'phpunit';
         $batch = $phpunit . '.bat';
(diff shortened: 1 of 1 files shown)


julianseeger:
Detecting the OS via directory separator sounds quite hacky to me (even though it probably works).
Wouldn't it be nicer to check truly os-specific things like `stripos(getenv('OS'), 'WINDOWS') !== false`?

porebskk:
You are right, but Smyfony Process does the same kind of detection, but I cant seem to find the reasoning behind this kind of specific check. https://github.com/symfony/process/blob/ba61437eff9cbdc6a29e469e3aad687e0053b214/Process.php#L277

Also check out https://stackoverflow.com/questions/5879043/php-script-detect-whether-running-under-linux-or-windows

I would argue that checking the DIRECTORY_SEPARATOR is more performance-optimized, but I understand the concern of the solution would be too hacky. You as collaborator may decide, I would change the PR and try to squash the commits once done.

julianseeger:
Well the symfony code is an argument. As paratest depends on symfony, we don't need to be more religious than that.

Thanks for the contribution. You will have to use dev-master to get that patch until the packagist integration is working again.

PR #296, opened 2018-01-25
Fix handling of special characters in JUnit XML

jeffchan:
PR #127 started escaping all HTML entities in order to produce valid JUnit XML. However, this is problematic because characters like `—` (`&mdash;`) and `…` (`&hellip;`) get escaped and produce invalid XML due to the `&` character.

Switching to `htmlspecialchars` ensures that only intended characters are escaped.

### DIFF
--- a/src/Logging/JUnit/Writer.php
+++ b/src/Logging/JUnit/Writer.php
@@ -164,7 +164,7 @@ class Writer
     protected function appendDefects(\DOMElement $caseNode, array $defects, string $type)
     {
         foreach ($defects as $defect) {
-            $defectNode = $this->document->createElement($type, htmlentities($defect['text']) . "\n");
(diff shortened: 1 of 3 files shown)


asuth:
is that double_encode param not useful here?

quizlet-chiragmedhora:
Thank you, @jeffchan !

jeffchan:
@julianseeger Yep. I just added. Without the fix, the test fails:
```
There was 1 error:

1) ParaTest\Logging\JUnit\WriterTest::testDataProviderWithSpecialCharacters
PHPUnit_Framework_Exception:
Entity 'mdash' not defined

Entity 'mdash' not defined


/opt/projects/paratest/test/unit/ParaTest/Logging/JUnit/WriterTest.php:48

```

PR #311, opened 2018-03-19
#310 Implement new SqliteRunner which uses sqlite to dispatch tests.

### LINKED ISSUE #310: paratest hangs indefinitely in some cases
Hi there,

first up: Thanks for devoting so much time into parallelizing PHPUnit. It's saved us much time during deployments and we wouldn't want to miss it.

Sadly, we're experiencing some non-reproducible issues during some runs of paratest in which the process just hangs indefinitely after the last test. We tried debugging it using strace and found that it keeps looping through a `select()` syscall. The strace output starting it looks as follows:
```
4857  exit_group(0)                     = ?
4857  +++ exited with 0 +++
4852  <... select resumed> )            = ? ERESTARTNOHAND (To be restarted if no handler)
4852  --- SIGCHLD {si_signo=SIGCHLD, si_code=CLD_EXITED, si_pid=4857, si_uid=126003, si_status=0, si_utime=40230, si_stime=136} ---
4852  select(8, [7], [], [], {0, 823288}) = 0 (Timeout)
4852  select(8, [7], [], [], {1, 0})    = 0 (Timeout)
4852  select(8, [7], [], [], {1, 0})    = 0 (Timeout)
```
The last line then repeats every second until the end of time.

We tried following the program path and found that the error may occur in `WrapperRunner::waitForStreamsToChange()`, since this is where the `select()` syscall is being made. Our theory is that one worker quits without letting the parent know and then the parent thinks there is still one worker left, but we don't have enough insight in how paratest works to debug it any further. Maybe you can give us a hint or any guess on what might happen, so we can try to develop a fix.

Debug information:
* This can be reproduced on a Debian and CentOS VM, but not locally on macOS.
* Paratest hangs in about 1 in 4 runs.
* All machines have enough memory built in so they don't have to swap.
* We're running 447 tests with 2312 assertions in 4 processes on 4 cores with the `WrapperRunner` runner and no xdebug extension enabled.
* The error occurs on both PHP 7.1 and 7.2. Versions below that haven't been tested.
* We've been using paratest version 1.1 with PHPUnit 6 as well as the master branch with PHPUnit 7.

If you like, we can send you the strace output via email.

Thanks for your help!

KarimGeiger:
This commit implements a SqliteRunner which attempts to resolve the
issues faced in #310 with the WrapperRunner. It's derived from the
WrapperRunner and dispatches n workers upon start, which then
independently fetch pending tests from a pre-populated sqlite database
until all tests are completed.

This commit also changes the way the runners are chosen during
bootstrapping. It will now dynamically check for the class in the
default namespace.

In addition, this commit will remove the isRunning flag from the
WrapperRunner because accessing the process information is more
reliable.

closes #310


bin/phpunit-sqlite-wrapper
```diff
@@ -0,0 +1,36 @@
...
+if (file_exists(__DIR__ . '/../vendor/autoload.php')) {
+    require_once __DIR__ . '/../vendor/autoload.php';
+}
+// Composer installation
+if (file_exists(__DIR__ . '/../../../autoload.php')) {
+    require_once __DIR__ . '/../../../autoload.php';
+}
+
+while ($test = $db->query('SELECT id, command FROM tests WHERE reserved_by_process_id IS NULL ORDER BY file_name LIMIT 1')->fetch()) {
+    $db->prepare('UPDATE tests SET reserved_by_process_id = :procId WHERE id = :id')
```
michaelbutler: To prevent race conditions, after running this `UPDATE` query you should check the affected row count result. If affected rows is `0`, that means another process has already successfully run the UPDATE query for the same test, and you should do a `continue;` to try grabbing another set with `SELECT`.

michaelbutler: Oh, and also in the UPDATE query you should have `WHERE id = :id AND reserved_by_process_id IS NULL` in case another process already updated that column.

michaelbutler:
This is cool! Not sure why the build failed just on that one instance, maybe you can try it out with PHP 7.1?

KarimGeiger:
Thanks for the feedback. Certainly a good idea to avoid race conditions, will updated the PR tomorrow.

We couldn’t really figure out why it failed there either. The curious part is that it worked for the other 7.1 targets and even the other tests on the same target. Maybe the reason for it failing was the bug you mentioned above. We’ll look into it.

KarimGeiger:
The race condition prevention fix is now ready and all tests are running through smoothly. Let me know if there's something else we should look into :)

michaelbutler:
Nice work. I'll test it out on some huge projects (https://www.exakat.io/largest-php-applications-2018/) and compare results 😃

michaelbutler:
Well turns out I can't run it on any of those huge projects due to not being a compatible phpunit version 😅

KarimGeiger:
Oh, damn 🤭
Well, if it helps, our project runs fine with it. Maybe we find some volunteers in this thread \*looks around*

michaelgv:
Looks good, merged. Thanks @KarimGeiger

KarimGeiger:
Thanks! You're very welcome. We had fun developing it 👍

PR #318, opened 2018-05-14
support phpunit setting environment variables

### LINKED ISSUE #103: ENV variables not passed to phpunit tests using standard runner (linux)
We use Jenkins to set a number of ENV variables to convey configuration into our (Selenium) PHPunit testcases.

We use paratest (0.7.2) to parallelize our testcases.

For some strange reason, we don't get our ENV variables in our testcase when using the default Paratest runner. It causes our Selenium PHPUnit testcases to fail since we miss config.

However, when using --runner WrapperRunner , the env variables are passing to our testcases, but that WrapperRunner implementation is not stable for our testsuite ("Worker has crashed" error when test failure occurs)

PHP 5.4.18
PHPUNIT 3.7.37
Linux SLES 11SP2  x64
Jenkins on JDK8
Symfony 2.3

Paratest invoked with : 

--configuration phpunit-selenium.xml.dist
--processes 8
--colors
--bootstrap '/mywebapp/src/app/bootstrap.php.cache'
--log-junit '/data/phpunit-selenium-testresults.xml'

Any thoughts ? 
We have had this working before, but for some reasons it no longer is ...

BePsvPT:
This pull request try to fix #103. I have tested in my personal project and it works. However, I'm not sure it is eventually fixed or not.

### DIFF
--- a/phpunit.xml.dist
+++ b/phpunit.xml.dist
@@ -24,4 +24,9 @@
             <directory suffix=".php">src</directory>
         </whitelist>
     </filter>
+
+    <php>
+        <env name="APP_ENV" value="testing"/>
+        <env name="CACHE_DRIVER" value="array"/>
+    </php>
 </phpunit>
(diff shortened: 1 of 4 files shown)


andreasschroth:
Thank you for your contribution, looks good... I merged it.

Issue #324, opened 2018-06-01
paratest does not work in its own project by default


mmeuser:
We have a problem running paratest. It aborts without running phpunit at all:

```
Running phpunit in 5 processes with C:\Users\m.meuser\phpstorm\paratest-fork\vendor\bin\phpunit.bat

Configuration read from C:\Users\m.meuser\phpstorm\paratest-fork\phpunit.xml.dist


In ResultPrinter.php line 198:

  Log file C:\Users\m.meuser\AppData\Local\Temp\PT_1A78.tmp is empty. This means a PHPUnit process has crashed.

The process: C:\php\php-7.2.2-nts-Win32-VC15-x64\php.exe C:\Users\m.meuser\phpstorm\paratest-fork\vendor\bin\phpunit.bat --configuration C:\Users\m.meuser\phpstorm\paratest-fork\phpunit.xml.dist --log-junit C:\Users\m.meuser\AppData\Local\Temp\PT_1A78.tmp ParaTest\Console\Commands\ParaTestCommandTest C:\Users\m.meuser\phpstorm\paratest-fork\te  st\unit\Console\Commands\ParaTestCommandTest.php  
This means a PHPUnit process was unable to run "C:\Users\m.meuser\phpstorm\paratest-fork\test\unit\Console\Commands\ParaTestCommandTest.php"


paratest [-p|--processes PROCESSES] [-f|--functional] [--no-test-tokens] [-h|--help] [--coverage-clover COVERAGE-CLOVER] [--coverage-html COVERAGE-HTML] [--coverage-php COVERAGE-PHP] [--coverage-text] [-m|--max-batch-size MAX-BATCH-SIZE] [--filter FILTER] [--whitelist WHITELIST] [--phpunit PHPUNIT] [--runner RUNNER] [--bootstrap BOOTSTRAP] [-c|--configuration CONFIGURATION] [-g|--group GROUP] [--exclude-group EXCLUDE-GROUP] [--stop-on-failure] [--log-junit LOG-JUNIT] [--colors] [--testsuite [TESTSUITE]] [--path PATH] [--] [<path>]
```

It seems as if paratest passes the phpunit.bat File to the php.exe process. I guess the default phpunit binary causes the problem. Running the tests with the command `--phpunit vendor/phpunit/phpunit/phpunit` works fine.

mmeuser:
here is the log from the running but broken paratest on the project itself:

```
[...]

PR #327, opened 2018-06-04
#324 fixed - derivation of phpunit runner pointing to php instead of binary

mmeuser:
before:
used the composer shell script to run phpunit

now:
uses the phpunit php start script to run phpunit

### DIFF
--- a/test/bootstrap.php
+++ b/test/bootstrap.php
@@ -17,7 +17,6 @@ if (defined('PHP_WINDOWS_VERSION_BUILD')) {
 define('PARA_BINARY', $pb);
 define('PARATEST_ROOT', dirname(__DIR__));
 
-//check for .bat first if on windows.
 $phpunit_path = PARATEST_ROOT . DS . 'vendor' . DS . 'phpunit' . DS . 'phpunit' . DS . 'phpunit';
 define('PHPUNIT', $phpunit_path);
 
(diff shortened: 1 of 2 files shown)


michaelgv:
Looks good! Can you resolve the conflicts?

mmeuser:
Conflicts are resolved.

test/bootstrap.php
```diff
@@ -18,10 +18,7 @@
 define('PARATEST_ROOT', dirname(__DIR__));
 
 //check for .bat first if on windows.
```
andreasschroth: From what I can see, you don't check for the phpunit.bat file anymore on Windows, so this comment is outdated and should be removed.

andreasschroth:
Could you please add an explanation of the purpose of this change? What advantages does this one have over using the bin via `vendor/bin`? Now the folder of phpunit is hard-coded, let's say (probably not going to happen, but just an example), Sebastian changes the phpunit namespace from `phpunit` to `sebastianbergmann`, then this is going to fail.

andreasschroth:
Is this related to #332 ?

andreasschroth:
Okay, I can now see it's related to your issue #324 ... Well, if using phpunit directly fixes the Windows problem, then this is a great PR. 👍 

Thank you for your contribution, I'm going to merge this now. Next time, it would only be great if you could provide some more context to the PR or link to a specific issue.

andreasschroth:
Okay, I can now see it's related to your issue #324 ... I didn't see it at first, because it was only included in the PR title and not in the description. Well, if using phpunit directly fixes the Windows problem, then this is a great PR. 👍 

Thank you for your contribution, I'm going to merge this now.

PR #328, opened 2018-07-09
Small tweaks

carusogabriel:
Together with PHPStorm and https://github.com/kalessil/phpinspectionsea, here are some improvements that we can make to Paratest, such as unnecessaries parentheses; null coalesce operator usage; in_array instead of array_search + comparison, etc


andreasschroth:
Thanks, looks good. 👍

PR #336, opened 2018-09-22
[Feature] Allow setting processes to number of CPU cores

andrerom:
Makes it a bit more portable for documentation use, or scripting it (composer scripts, shell, ..).

I'm alos suggesting setting this as a default, however that is not a must here and up to what you see as best default for the project.

### DIFF
--- a/src/Runners/PHPUnit/Options.php
+++ b/src/Runners/PHPUnit/Options.php
@@ -116,6 +116,8 @@ class Options
 
         if ($opts['processes'] === 'auto') {
             $opts['processes'] = $this->getNumberOfCPUCores();
+        } elseif ($opts['processes'] === 'half') {
+            $opts['processes'] = intdiv($this->getNumberOfCPUCores(), 2);
         }
 
         $this->processes = $opts['processes'];
--- a/src/Runners/PHPUnit/Options.php
+++ b/src/Runners/PHPUnit/Options.php
@@ -115,9 +115,9 @@ class Options
         }
 
         if ($opts['processes'] === 'auto') {
-            $opts['processes'] = $this->getNumberOfCPUCores();
+            $opts['processes'] = self::getNumberOfCPUCores();
         } elseif ($opts['processes'] === 'half') {
-            $opts['processes'] = intdiv($this->getNumberOfCPUCores(), 2);
+            $opts['processes'] = intdiv(self::getNumberOfCPUCores(), 2);
         }
 
         $this->processes = $opts['processes'];
@@ -334,8 +334,10 @@ class Options
      * Return number of (logical) CPU cores, use 2 as fallback.
      *
      * Used to set number of processes if argument is set to "auto", allows for portable defaults for doc and scripting.
+     *
+     * @internal
      */
-    private function getNumberOfCPUCores(): int
+    public static function getNumberOfCPUCores(): int
     {
         $cores = 2;
         if (is_file('/proc/cpuinfo')) {
--- a/src/Runners/PHPUnit/Options.php
+++ b/src/Runners/PHPUnit/Options.php
@@ -114,6 +114,10 @@ class Options
             $opts[$opt] = $opts[$opt] ?? $value;
         }
 
+        if ($opts['processes'] === 'auto') {
+            $opts['processes'] = $this->getNumberOfCPUCores();
+        }
+
         $this->processes = $opts['processes'];
         $this->path = $opts['path'];
         $this->phpunit = $opts['phpunit'];
@@ -180,7 +184,7 @@ class Options
     protected static function defaults(): array
     {
         return [
-            'processes' => 5,
+            'processes' => 'auto',
             'path' => '',
             'phpunit' => static::phpunit(),
             'functional' => false,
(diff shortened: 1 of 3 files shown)


andreasschroth:
Thank you for your PR. Could you please the failing tests (see TravisCI)?

src/Runners/PHPUnit/Options.php
```diff
@@ -113,6 +113,10 @@ public function __construct(array $opts = [])
         foreach (self::defaults() as $opt => $value) {
             $opts[$opt] = $opts[$opt] ?? $value;
         }
+        
+        if ($opts['processes'] === 'auto') {
+            $opts['processes'] = $this->getNumberOfCPUCores();
```
Nemo64: I'm not sure but I think there should be at least 2 processes since that helps even on a single core machine and some build pipelines have just a single core assigned.

andreasschroth: I was also thinking about the best value when reviewing the PR. But we also want to avoid that the system starts lacking, etc. What value per core do you use?

andrerom: Personally I use cores - 1

So one is left for parent, also to not freeze the machine

Nemo64: Machine freezing is more of an io issue on linux than cpu resources. There is `ionice` for that.

But I see that there are a lot of cases. I wouldn't overload hyperthreads while I'd definitly do that if I only had real cores. My idea of "at least 2" stems from build pipelines with 1 virtual core. Also It wouldn't make much sense to run paratest with 1 process. But I also admit that I could just define 2 processes in my build pipeline and for the local machine it's probably fine to have just as many processes as cores.

Since the default is now "auto" another feasible solution would be `100%` or `150%` to define over/underloading the machine.

BUT its probably best to leave it as is and merge this first and than think about a better algorithm if there is a need for it.

andrerom: We can make a special case when value is 1 and set it to 2.
Meaning two is minimum, this can make sense as it is also the default.

**Or** we ca change "auto" to introduce a set of strategies, either percentages as you mention or something ala:
- Cores
- Coresx2 ("hyper threading ceiling mode")
- Cores+1
- Cores-1

Where we pick on of them as default.
But I think I would prefer percentages to keep things "simple".

Either way this can make sense given there are as you mention many use cases.
If a given test suite is set of integration testing using things like cache service and maybe also full database then there are plenty of other processes already competing for available system resources.

andreasschroth: Not sure about this. I also thought yesterday about it and that it might be nice to set a dynamic value relative to the number of CPU cores. But it might already be too complicated and actually I think the setting to set the total number of processes is already absolutely sufficient and fine.

I mean, if someone has 4 cores and wants to use the strategy "cores x 2", he can just set it to 8. If someone has 10 cores and wants to use the strategy "cores + 1", he can just set it to 11 by himself. What I mean is, if someone sets the value manually anyway and changes the config, he already put thought into it. Having 2 settings for it then (like a strategy for the number of CPU cores and also the total number of processes) would just be a duplicate and imo, a useless complication.

So for me, I would just say that "auto" should provide a reasonable default. If you want to change it to something different, the developer should just set the number of processes accordingly and thus apply whatever strategy he likes.

And I also think "cores x 2" is a good default strategy - it shouldn't make the host lack, etc. Then, we can also skip the min value of 2, as every machine obviously has 1 core already, which leads to a min value of 2.

Nemo64: I like the idea. Just a simple selection. If you are using `paratest` you probably want your tests to be done fast so just default to process spam. If it annoys you, then adjust it manually. My build pipelines have 1 core so 2 is what I was using anyways but if I'd port my app somewhere else with 2 vcores, 4 is fine.

I did some testing with a project i'm working on. I have 2 cores 4 threads. Running the functional tests with 8 threads isn't much different than running them with 5 but it is definitely faster than running them with 4 threads.

The only concern with that solution is how well paratest handles massive amounts of processes like on a Threadripper with 32 cores and 64 threads. It would then run with 128 processes.

andrerom: > The only concern with that solution is how well paratest handles massive amounts of processes like on a Threadripper with 32 cores and 64 threads. It would then run with 128 processes.

As far as I understand @andreasschroth it will be 64, cores (32) x2.  At least the intention for the code here is to count the physical cores and not the virtual once.

src/Runners/PHPUnit/Options.php
```diff
@@ -327,4 +331,35 @@ private function isFile(string $file): bool
...
+            preg_match_all('/^processor/m', $cpuinfo, $matches);
+            $cores = count($matches[0]);
+        } elseif (DIRECTORY_SEPARATOR === '\\') {
+            // Windows
+            if (($process = @popen('wmic cpu get NumberOfCores', 'rb')) !== false) {
+                fgets($process);
+                $cores = (int) fgets($process);
+                pclose($process);
+            }
+        } elseif (($process = @popen('sysctl -a', 'rb')) !== false) {
```
Nemo64: you can just query the number by using `sysctl -n hw.ncpu`. No need to parse the output

src/Runners/PHPUnit/Options.php
andreasschroth: @Nemo64: I believe (didn't test it myself though) that there shouldn't be any problems. And not sure if the code from @andrerom retrieves the number of physical cores or the logical ones. But normally for the OS there's no difference, as it only works with the logical cores usually. At least, that's what I know. So yes, I agree with you.

But setting up some DigitalOcean droplet for example for 1 hour and running some tests on 16 or 24 vCPUs shouldn't be a big deal to test it.

andrerom: So summary, what should be changed here?

The core count x2 when set to auto?

andrerom:
@andreasschroth PR updated

One missing piece is updating help text of command, but maybe it's better if you do that?

The new modes, added one given the extensive discussions on sane defaults above to cover most needs:
- 'auto' _(default)_: _Meant to run number of processes equal to number of logical cores to maximize test execution for most common cases_
- 'half': _Runs number of processes equal to half of number of logical cores, leaving CPU cable of handling additional processes in addition to php processes like database, search, cache, ... in case of integration tests._

src/Runners/PHPUnit/Options.php
andreasschroth: @andrerom Thanks. I'm just missing the min value of 2, if the server just has 1 core... Otherwise paratest wouldn't serve any purpose.

andreasschroth:
@andrerom Thanks, will update documentation.

andreasschroth:
@andrerom Thank you very much. I merged your PR and updated the documentation. I plan to tag a new version `2.1` of paratest later today or tomorrow.

andreasschroth:
Alright, Version 2.1 is tagged & released. Also available as .phar now.

PR #322, opened 2018-05-31
Publish Phar executable on Release pages

Gasol:
To resolve #176, The Phar executable is very useful when root project has some dependency conflict with paratest, for example `symfony/console`.

This PR require Github token to work properly, You can replace APIKEY by editing files (**"Allow edits from maintainers has checked"**), It should work like following link.
https://github.com/Gasol/paratest/releases/tag/2.0.0

See also  https://docs.travis-ci.com/user/deployment/releases/#Authenticating-with-an-OAuth-token and https://github.com/travis-ci/travis-ci/issues/2457#issuecomment-47113151

PR #338, opened 2018-09-26
Add deploy section to .travis.yml.

andreasschroth:
Unfortunately, the deploy section was missing in PR #322 , thus this PR.


andreasschroth:
I hope it works. We should see it the next time when a new release is tagged, as the deploy only happens when the commit is tagged as a release.

PR #345, opened 2018-10-19
Change default console argument to auto

andrerom:
I forgot this part in prior pull request, so doc and command is not aligned now in 2.1.

### DIFF
--- a/src/Console/Commands/ParaTestCommand.php
+++ b/src/Console/Commands/ParaTestCommand.php
@@ -39,7 +39,7 @@ class ParaTestCommand extends Command
     protected function configure()
     {
         $this
-            ->addOption('processes', 'p', InputOption::VALUE_REQUIRED, 'The number of test processes to run.', 5)
+            ->addOption('processes', 'p', InputOption::VALUE_REQUIRED, 'The number of test processes to run.', 'auto')
             ->addOption('functional', 'f', InputOption::VALUE_NONE, 'Run methods instead of suites in separate processes.')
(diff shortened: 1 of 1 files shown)


andreasschroth:
Oops! Thank you. But could you please fix the failing tests again? TravisCI is failing...

andreasschroth:
@andrerom Would be awesome, if you could update the failing tests. Otherwise I'll do that.

andrerom:
I won't be able to anytime soon.

andreasschroth:
@andrerom Thanks. Well I guess phpunit changed something in their method signature, is definitely unrelated to this PR.

Regarding the 32 cores I was also a bit surprised at first - and thought they might be extremely generous. But yes, I looked it up under https://docs.travis-ci.com/user/reference/overview/ and it seems that Travis VMs run on 2 cores. I find it a bit weird though that `/proc/cpuinfo` inside the VM contains the information about the host, how exactly can that be?

andrerom:
> how exactly can that be?
🤷‍♂️ 

Can't remember seeing this when testing on a ubuntu docker container locally long time ago, but as this is completely different hypervisor I'm unsure if it matters or not.

matthewnessworthy:
The unrelated failure relates to this commit ( https://github.com/sebastianbergmann/phpunit/commit/f5e5add13e73933c65878f53d5a94cdc0f120cc4#diff-511c90242d5589e941582006e0485184R407 ) from 3 days ago

PR #350, opened 2018-11-22
Travis test - pin max phpunit version to ^7.5@dev instead of *@dev

matthewnessworthy:
The phpunit TestCase setUp signature has changed in the latest master branch of phpunit in the commit https://github.com/sebastianbergmann/phpunit/commit/f5e5add13e73933c65878f53d5a94cdc0f120cc4#diff-511c90242d5589e941582006e0485184R407

Pin the highest tested version of phpunit to ^7.5@dev

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -31,7 +31,7 @@ matrix:
 
 install:
   - if [[ $CS_CHECK == 'true' ]]; then phpenv config-rm xdebug.ini || return 0; else composer remove --dev --no-update --no-scripts friendsofphp/php-cs-fixer; fi;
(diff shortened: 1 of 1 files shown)


andreasschroth:
Thanks for your PR, but I don't really want to merge it as for me it seems like the wrong solution to the problem. We better fix our code and make the signature match to the latest phpunit changes - that's what the CI tests are for, that we can see if Paratest still works with the latest phpunit dev versions or not - if we just set the phpunit dev version to a fixed one, it doesn't fix anything but is more like a trick.

PR #351, opened 2018-11-22
--parallel-suites documentation & path fix

matthewnessworthy:


src/Runners/PHPUnit/FullSuite.php
```diff
@@ -21,7 +21,7 @@ class FullSuite extends ExecutableTest
      */
     public function __construct($suiteName, $configPath)
     {
-        parent::__construct($suiteName);
+        parent::__construct('');
```
andreasschroth: What is this change for?

andreasschroth:
Thank you for your contribution, what is the change in the constructor call for?

matthewnessworthy:
the constructor was incorrectly a double path

andreasschroth:
@matthewnessworthy Yes, right, thanks for the fix and the added documentation. 👍 

PS: Greetings to Gozo from St. Pauls Bay.

matthewnessworthy:
👋 what are the chances :)

Slamdunk:
Test added in https://github.com/paratestphp/paratest/pull/494

PR #361, opened 2019-01-24
Adds --passthru and --passthru-php options. Adds --verbose option.

### LINKED ISSUE #360: Add --passthru and --passthru-php options
I propose to add two options to `paratest`

````
--passthru        Pass the given arguments verbatim to the underlying test framework. Example: --passthru="'--prepend xdebug-filter.php'"
--passthru-php    Pass the given arguments verbatim to the underlying php process. Example: --passthru-php="'-d' 'zend_extension=xdebug.so'"
````

## --passthru  
Paratest has the inherent problem of being tightly couple to the underlying test framework (e.g. phpunit). Phpunit is under active development and gets reguarly new features, e.g. the `--prepend` option for better/faster code coverage generation (see https://thephp.cc/news/2019/01/faster-code-coverage ).

We would now need an explicit PR on paratest to get this functionality. By supporting a generating "just pass everything I give to you to phpunit", the end user can make use of the new feature immediately.
This **does not** mean that paratest should not support a new feature in a stable (and tested) way, but it makes it much easier to adopt paratest.

## --passthru-php
The same functionality could be used to make php options availale (e.g. as requested in https://github.com/paratestphp/paratest/issues/354)

## Usage example
`bin/paratest --runner WrapperRunner --passthru="'--prepend' 'xdebug-filter.php'" --passthru-php="'-d' 'zend_extension=xdebug.so'"`

The individual runners would then be invoked with something like
`/usr/bin/php7.2 '-d' 'zend_extension=xdebug.so' '/codebase/paratest/vendor/phpunit/phpunit/phpunit' '--configuration' '/codebase/paratest/phpunit.xml.dist' '--log-junit' '/tmp/PT_uXvYTy' 'ParaTest\Util\StrTest' 'test/unit/Util/StrTest.php' '--prepend xdebug-filter .php'"`

## Note
- Providing proper escaping is the responsibility of the user, i.e. the user has to take care that what ever is passed in `--passthru` or `--passthru-php` is properly escaped. Individual arguments should always be embedded in single quotes to avoid issues, e.g.

```
--passthru="'--prepend' 'xdebug-filter.php'"
```
instead of 

```
--passthru="--prepend xdebug-filter.php"
```

### PR DESCRIPTION
**Pull request by paslandau**
Fixes #360 and #354

### CODE REVIEW

File: src/Runners/PHPUnit/ExecutableTest.php [hunk 1]
```diff
@@ -175,23 +175,36 @@ public function getLastCommand(): string
...
+     * @param string      $binary
+     * @param array       $options
+     * @param array       $environmentVariables
+     * @param string|null $passthru
+     * @param string|null $passthruPhp
      *
      * @return $this
      */
-    public function run(string $binary, array $options = [], array $environmentVariables = [])
+    public function run(string $binary, array $options = [], array $environmentVariables = [], string $passthru = null, string $passthruPhp = null)
```
andreasschroth: Imo, the param type declaration needs to be `?string`, otherwise it wouldn't be possible to pass (theoretically, when method is called in PHP code) `null` as `$passthru` and then set `$passthruPhp`.

src/Runners/PHPUnit/ExecutableTest.php
```diff
@@ -201,6 +214,33 @@ public function run(string $binary, array $options = [], array $environmentVaria
...
+     * php -Dzend_extension=xdebug.so vendor/bin/phpunit --teststuite suite1 --prepend xdebug-filter.php.
+     *
+     * @param $binary
+     * @param $options
+     * @param string|null $passthru
+     * @param string|null $passthruPhp
+     *
+     * @return string
+     */
+    protected function getFullCommandlineString($binary, $options, string $passthru = null, string $passthruPhp = null)
```
andreasschroth: Same here, I think param type needs to be `?string` for `$passthru` and `$passthruPhp`.

src/Runners/PHPUnit/ExecutableTest.php
```diff
@@ -214,17 +254,20 @@ public function getToken(): int
     /**
      * Generate command line with passed options suitable to handle through paratest.
      *
-     * @param string $binary  executable binary name
-     * @param array  $options command line options
+     * @param string $binary   executable binary name
+     * @param array  $options  command line options
+     * @param null   $passthru
```
andreasschroth: I guess this should be `string|null`?

src/Runners/PHPUnit/ExecutableTest.php
```diff
@@ -305,15 +348,29 @@ protected function prepareOptions(array $options): array
...
      * @return mixed
      */
-    protected function getCommandString(string $binary, array $options = [])
+    protected function getCommandString(string $binary, array $options = [], string $passthru = null)
     {
         // The order we add stuff into $arguments is important
         $arguments = [$binary];
+        // Note:
+        // the arguments MUST come last and we need to "somehow"
+        // mere the passthru string in there.
```
andreasschroth: Typo? Should be "merge", right?

test/unit/Runners/PHPUnit/ExecutableTestTest.php
```diff
@@ -31,6 +33,26 @@ public function testGetCommandStringIncludesOptions()
...
+        $options = ['bootstrap' => 'test/bootstrap.php'];
+        $binary = '/usr/bin/phpunit';
+        $passthru = "'--prepend' 'xdebug-filter.php'";
+        $passthruPhp = "'-d' 'zend_extension=xdebug.so'";
+
+        $command = $this->call($this->executableTestChild, 'getFullCommandlineString', $binary, $options, $passthru, $passthruPhp);
+        // Note:
+        // '--log-junit' '/tmp/PT_LKnfzA'
+        // is appended by default where PT_LKnfzA is randomly generated - so we remove it from the resulting command
+        $command = preg_replace("# '--log-junit' '[^']+?'#", '', $command);
```
andreasschroth: OK nice, very strict way to test the result. 👍 (Personally I guess I would have just used `assertContains` here.)

paslandau: I would actually like to make the `'--log-junit' '/tmp/PT_LKnfzA` part configurable in `ExecutableTest` but that would probably be a bigger refactoring :)

As for "strictness": I try to avoid "contains" as it opens up the assertion for "unwanted" side effects + in this case I would need two assertions instead of one.

paslandau:
@andreasschroth

If this gets merged I would add a small write-up to the docu on how to get code code coverage running with paratest:
- preparing the phpunit.xml (incorrect whiltelist is often a problem)
- using php dbg
- using xdebug
- using phpunit's --prepend filter

andreasschroth:
Thanks a lot @paslandau  for the PR, perfect now! Got merged.

Better docs would be awesome as I can see some devs having problems how to get it properly set up. 

If you plan further PRs with new features, please let me know, otherwise I would tag a new version soon (maybe after the docs write-up?).

paslandau:
@andreasschroth docs done => https://github.com/paratestphp/paratest/pull/362 located under docs/code-coverage.md

Should be the last one for now - would appreciate a new version!

andreasschroth:
@paslandau Thank you! Awesome work. I just tagged Version 2.2.0, so everyone can make use of the latest improvements.

PR #366, opened 2019-02-03
Add phpunit8 support

Tlapi:


andreasschroth:
Thank you very much for your PR. 👍 

Could you please have a look at failing CI?

pekral:
I think, problem with PR will be resolved when this (https://github.com/sebastianbergmann/php-code-coverage/commit/63d3b21b3be6ab291a334838d210862ce4bc92ca) will be merged.

Tlapi:
Seems like this has been already fixed in master, so we are just waiting for a new release tag.

jdufresne:
You'll need to drop PHP 7.1 from Travis as PHPUnit 8 does not support it. You can then update the `php` require line `composer.json` to `"php": "^7.2",` (same as PHPUnit's)

jdufresne:
A new version of php-code-coverage was released today :tada:, so I believe this work can continue.

https://github.com/sebastianbergmann/php-code-coverage/releases/tag/7.0.2

Tlapi:
Ok, I did some more work. Im not sure why PHPUNIT_DEV=true is failing, but the rest looks ok.

jdufresne:
Can you drop PHP 7.1 from the travis configuration? It is no longer supported.

andreasschroth:
Thank you, looks good. 👍Merged.

jdufresne:
> Ok, I did some more work. Im not sure why PHPUNIT_DEV=true is failing, but the rest looks ok.

I looked into this a bit. `PHPUnitWarningsTest` is failing because on dev because one must pass `--fail-on-warnings` for `phpunit` to fail when only warnings occur. The subprocess does not do this, so it has an exit code of 0 instead of 1.

I bisected this to the following phpunit commit:

https://github.com/sebastianbergmann/phpunit/commit/d00d0af031bdb4dc97fcfffb9088825ed86772eb

And issue:

https://github.com/sebastianbergmann/phpunit/issues/3531

PR #371, opened 2019-02-16
Fix PHPUnitWarningsTest with PHPUnit @dev

jdufresne:
A PHPUnit bug was fixed to not fail on a warning by default. To test
failures on warnings, define this in the test fixtures XML configuration
file.


andreasschroth:
Thank you, merged! 👍

PR #372, opened 2019-02-16
Add .phpunit.result.cache to gitignore

jdufresne:
The optional caching of test results that was introduced in PHPUnit 7.3
is now enabled by default in PHPUnit 8. This creates a
.phpunit.result.cache file.


andreasschroth:
Thank you, merged.

PR #376, opened 2019-02-25
WrapperRunner: proper PHPUNIT_COMPOSER_INSTALL constant initialization

crzdeveloper:
## Problem
when I run paratest with enabled [Logging WrapperRunner output](https://github.com/paratestphp/paratest/blob/master/docs/logging.md), I got a strange error:
```
PHPUnit\Framework\Exception: Fatal error: Uncaught Error: Class 'PHPUnit\Util\Configuration' not found in - on line 425

Error: Class 'PHPUnit\Util\Configuration' not found in - on line 425

Call Stack:
    0.0013     549496   1. {main}() -:0
```
## Info
```
phpunit/phpunit: 7.2.0
brianium/paratest: 2.2.0
PHP version: 7.1
```
## How to reproduce
In our project I just followed this documentation: [Logging WrapperRunner output](https://github.com/paratestphp/paratest/blob/master/docs/logging.md) and it always results in the described error.
## Investigation
When the paratest is run with WrapperRunner enabled, the phpunit constant PHPUNIT_COMPOSER_INSTALL is not set. When the code execution reaches the following piece of code in the ```\PHPUnit\Framework\TestCase::run``` https://github.com/sebastianbergmann/phpunit/blob/7.2.0/src/Framework/TestCase.php#L737 :
```php
if (\defined('PHPUNIT_COMPOSER_INSTALL')) {
    $composerAutoload = \var_export(PHPUNIT_COMPOSER_INSTALL, true);
} else {
    $composerAutoload = '\'\'';
}
```
the condition is false. Later, when the template https://github.com/sebastianbergmann/phpunit/blob/7.2.0/src/Util/PHP/Template/TestCaseClass.tpl.dist#L20 is assembled and executed, this piece of code does nothing because the ```$composerAutoload``` is empty:
```php
if ($composerAutoload) {
    require_once $composerAutoload;
    define('PHPUNIT_COMPOSER_INSTALL', $composerAutoload);
} else if ($phar) {
    require $phar;
}
```
And later this line https://github.com/sebastianbergmann/phpunit/blob/7.2.0/src/Util/PHP/Template/TestCaseClass.tpl.dist#L84 results in the error
```
PHPUnit\Framework\Exception: Fatal error: Uncaught Error: Class 'PHPUnit\Util\Configuration' not found in - on line 425
```
(the exact line in your environment will vary)

## Further thoughts
Would be great if anybody can reproduce and test this PR in their environments. I found this problem in the paratest v2.2.0, but it should be reproducible in the latest version of the praratest.

### DIFF
--- a/bin/phpunit-wrapper
+++ b/bin/phpunit-wrapper
@@ -1,13 +1,19 @@
 <?php
 echo "Worker starting\n";
 
-// git working copy
-if (file_exists(__DIR__ . '/../vendor/autoload.php')) {
-    require_once __DIR__ . '/../vendor/autoload.php';
-}
-// Composer installation
-if (file_exists(__DIR__ . '/../../../autoload.php')) {
-    require_once __DIR__ . '/../../../autoload.php';
+$composerAutoloadFiles = [
+    __DIR__ . '/../../../autoload.php',
+    __DIR__ . '/../../vendor/autoload.php',
+    __DIR__ . '/../vendor/autoload.php'
+];
+
+foreach ($composerAutoloadFiles as $file) {
+    if (file_exists($file)) {
+        require_once $file;
+        define('PHPUNIT_COMPOSER_INSTALL', $file);
+
+        break;
+    }
 }
 
 $commands = fopen('php://stdin', 'r');


andreasschroth:
Thanks for your PR.

I'm alright with merging your PR as I can't see how it would do any harm, but I would be interested in your directory structure. So from what I can see your problem is that your autoloader lies in `__DIR__ . '/../../vendor/autoload.php'` as this is the one that hasn't been checked before.

The only question is, why is it there?

crzdeveloper:
Actually, I did that autoloader inclusion magic in the same way as it is done in the ```vendor/bin/phpunit```: https://github.com/sebastianbergmann/phpunit/blob/7.2.0/phpunit#L30

My logic was the following: the phpunit file tries to find the ```autoload.php``` in the following places:
```php
    __DIR__ . '/../../autoload.php',
    __DIR__ . '/../vendor/autoload.php',
    __DIR__ . '/vendor/autoload.php'
```
The ```phpunit-wrapper``` tries the following paths:
```php
__DIR__ . '/../vendor/autoload.php',
__DIR__ . '/../../../autoload.php'
```

So I decided that maybe one of the paths was forgotten and I added 
```php
__DIR__ . '/../../vendor/autoload.php'
```

Actually our project structure looks like this:
```
project/vendor/brianium/paratest/bin/phpunit-wrapper
project/vendor/autoload.php
```

So the new path that I added is not relevant to our project. Please, let me know if I need to remove that path. Thank you!

andreasschroth:
@crzdeveloper Thanks for explanation. I just merged your PR in order to keep it consistant to the way PhpUnit handles it. But imo, that ```__DIR__ . '/../../vendor/autoload.php``` path should never be required.

Issue #377, opened 2019-02-25
WrapperRunner: PHP session support


crzdeveloper:
## Problem
When I use the WrapperRunner in order to gather some logs, PHP can not initialize the session:

```
Zend_Session_Exception: Session must be started before any output has been sent to the browser;
output started in [.......]/vendor/brianium/paratest/bin/phpunit-wrapper/2 in
[.......]/vendor/zendframework/zendframework1/library/Zend/Session.php on line 456
```
It seems the output is started in https://github.com/paratestphp/paratest/blob/master/bin/phpunit-wrapper#L2
## Info
```
phpunit/phpunit: 7.2.0
brianium/paratest: 2.2.0
PHP version: 7.1
```
## Investigation
If I comment out that line in the ```bin/phpunit-wrapper```, the ```WrapperRunner``` stucks here: https://github.com/paratestphp/paratest/blob/master/src/Runners/PHPUnit/WrapperRunner.php#L72

P.S.: I know that starting the session in unit tests is not the best practice, but that's the legacy code :)

michaelbutler:
I don't think we can support sessions in the same process; understandably you are working with legacy code, but that is too wild of an edge case to update paratest for everyone.

I think a solution for you could be to just enable [session autostart](http://php.net/manual/en/session.configuration.php#ini.session.auto-start) in your PHP ini that you use for unit tests? That way the session will be automatically started before any PHP code is actually run. Note that this won't work right if the session contains php objects that have to be deserialized, but since you're doing unit tests there should be no pre-existing state.

crzdeveloper:
It is fine for me even if this issue will be closed. Anyway we wanted to refactor legacy session stuff.

If I understand the logic correctly, the parent process waits for the child process till it outputs anything. As another solution, the parent process could mark the newly created child as "ready-to-accept-commands" immediately after the creation by adding them into the ```$this->modified``` collection. Enough talking :) - here is a possible solution: #378

PR #378, opened 2019-02-26
WrapperRunner: prevent unnecessary output thus allow PHP session

crzdeveloper:
A possible solution for #377.

I did changes in paratest v2.2.0 (PHP 7.1) and then applied them onto the master branch. I don't have PHP 7.2 environment, so couldn't test it on master.

### DIFF
--- a/bin/phpunit-wrapper
+++ b/bin/phpunit-wrapper
@@ -1,5 +1,4 @@
 <?php
-echo "Worker starting\n";
 
 // git working copy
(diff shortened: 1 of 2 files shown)


michaelbutler:
I'd prefer to keep the existing functionality as-is, and instead add an option to "turn off output" which you could then enable via a pass through or ENV setting.

andreasschroth:
Thank you for your PR, but won't merge. I don't think this is the way to go.

I agree with @michaelbutler - would be better to provide a setting to turn off output.

PR #398, opened 2019-07-10
Allow iterable data providers

thewilkybarkid:
Data providers aren't always arrays, they can be any iterable (using Generators is a common pattern).


andreasschroth:
Thank you. 👍Makes sense. Merged.

PR #403, opened 2019-10-08
Fixed Symfony 4 deprecated Process constructor

alldroll:
This fixes the error:

>"Passing a command as a string when creating a "Symfony\Component\Process\Process" instance is deprecated since Symfony 4.2, pass it as an array of its arguments instead, or use the "Process::fromShellCommandline()" constructor


andreasschroth:
Thank you for your contribution! Looks good.

PR #411, opened 2019-11-12
Support multiple TestCase error reports with system out

### LINKED ISSUE #352: JUnit addSystemOut fails with warning when TestCase has multiple failures or errors
The following function attempts to process `<system-out>` from JUnit XML nodes produced by PHPUnit:

https://github.com/paratestphp/paratest/blob/87aa787eaf2f3d2d3baad14af50d18e277a56f6f/src/Logging/JUnit/TestCase.php#L129-L149

If the test case being processed contains more than one `<failure>` (or `<error>`), Paratest will throw out an ugly warning and stack trace during execution:

```
PHP Warning:  ParaTest\Logging\JUnit\TestCase::addSystemOut(): Cannot assign to an array of nodes (duplicate subnodes or attr detected) in /opt/hellofax/vendor/brianium/paratest/src/Logging/JUnit/TestCase.php on line 141
PHP Stack trace:
PHP   1. {main}() /opt/hellofax/vendor/brianium/paratest/bin/paratest:0
PHP   2. ParaTest\Console\ParaTestApplication->run($input = *uninitialized*, $output = *uninitialized*) /opt/hellofax/vendor/brianium/paratest/bin/paratest:35
```

Additionally, the system-out text won't appear anywhere in the generated JUnit XML output.

There are a couple things going wrong here, but I think the main takeaway is that Paratest shouldn't even be trying to concatenate the system-out text into the content of the failure and error node (or nodes) like this at all. Instead, this function shouldn't exist and the `caseFromNode` should be ingesting the node's `system-out` value directly (because it's supposed to be a child of the `<testcase>`). This will probably require adding a new property to the `ParaTest\Logging\JUnit\TestCase` class and some code in the `Writer` class to handle rendering it to the finished XML.

Fedcomp:
Added handling of phpunit xml output with multiple errors in single testcase.
```xml
<testsuites>
  <testsuite name="timeoutTest" file="/project/test/timeoutTest.php" tests="1" assertions="0" errors="2" warnings="0" failures="0" skipped="0" time="1.014721">
    <testcase name="testTimeoutNotice" class="timeoutTest" classname="timeoutTest" file="/project/test/timeoutTest.php" line="8" assertions="0" time="1.014721">
      <error type="PHPUnit\Framework\RiskyTestError">Risky Test
/project/vendor/phpunit/phpunit/src/TextUI/Command.php:200
/project/vendor/phpunit/phpunit/src/TextUI/Command.php:159
</error>
      <error type="PHPUnit\Framework\RiskyTestError">Risky Test
/project/vendor/phpunit/phpunit/src/TextUI/Command.php:200
/project/vendor/phpunit/phpunit/src/TextUI/Command.php:159
</error>
      <system-out>Custom error log on result test with multiple errors!</system-out>
    </testcase>
  </testsuite>
</testsuites> 
```
fixes #352 

I also added temporary files to .gitignore (they are generated when running tests locally). 

Unfortunately phpunit still don't show correct reason why test stopped in junit output. I will try to find/make issue in phpunit later.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -1,2 +1,4 @@
 vendor
 composer.lock
+/test/fixtures/generated-tests
+.phpunit.result.cache
--- a/src/Logging/JUnit/TestCase.php
+++ b/src/Logging/JUnit/TestCase.php
@@ -92,42 +92,6 @@ class TestCase
     }
 
     /**
-     * @param string $type
-     * @param string $text
-     */
-    public function addFailure(string $type, string $text)
-    {
-        $this->addDefect('failures', $type, $text);
-    }
-
-    /**
-     * @param string $type
-     * @param string $text
-     */
-    public function addError(string $type, string $text)
-    {
-        $this->addDefect('errors', $type, $text);
-    }
-
-    /**
-     * @param string $type
-     * @param string $text
-     */
-    public function addWarning(string $type, string $text)
-    {
-        $this->addDefect('warnings', $type, $text);
-    }
-
-    /**
-     * @param string $type
-     * @param string $text
-     */
-    public function addSkipped(string $type, string $text)
-    {
-        $this->addDefect('skipped', $type, $text);
-    }
-
-    /**
      * Add a defect type (error or failure).
      *
      * @param string $collName the name of the collection to add to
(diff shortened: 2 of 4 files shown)


andreasschroth:
Thank you a lot for your work. :+1:

andreasschroth:
As I think it's an important bugfix, I just released version 3.1.1.

Fedcomp:
Woah that's great, thank you!

PR #415, opened 2019-11-26
Add Crap4J coverage support

maxnk:


maxnk:
@andreasschroth, here is a problem: `\PHPUnit\Framework\TestCase::expectExceptionMessageRegExp` has been marked as deprecated since 8.0.0. And now, in 9.0-dev, replaced with another method.

Look like paratest needs major version bump for PHPUnit 9.

andreasschroth:
Apologies for the late reply. Thanks for your work, awesome! 👍 

I merged it now.

PR #422, opened 2020-02-07
Add PHPUnit 9 support

Slamdunk:
See:

1. https://github.com/sebastianbergmann/phpunit/blob/9.0.0/ChangeLog-9.0.md
1. https://github.com/paratestphp/paratest/pull/74
1. https://github.com/paratestphp/paratest/pull/241

### DIFF
--- a/src/Runners/PHPUnit/ExecutableTest.php
+++ b/src/Runners/PHPUnit/ExecutableTest.php
@@ -23,7 +23,7 @@ abstract class ExecutableTest
      * @var string
      */
     protected $temp;
-    protected $fullyQualifiedClassName;
+
     protected $pipes = [];
 
     /**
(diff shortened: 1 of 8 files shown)


src/Runners/PHPUnit/ExecutableTest.php
```diff
@@ -53,10 +53,9 @@ abstract class ExecutableTest
      */
     protected $lastCommand = '';
 
-    public function __construct(string $path, ?string $fullyQualifiedClassName = null)
```
Slamdunk: > Multiple test case classes (classes that extend TestCase) that are declared in a single sourcecode file are no longer supported

andreasschroth: Imo, a quite reasonable change from PHPUnit. :-)

Slamdunk:
I've tried this PR on our business SUT and it freeze at the end of the run.
100% the issue is not related to this PR but with one of the dozen new major release of PHPUnit and deps; still it must be pointed out.
On monday I'll try to dig further the problem.

andreasschroth:
@Slamdunk Thanks for your feedback. I'm also sure it's not related to this PR, thus I'm going to merge it once Travis shows green in a few minutes (had some network timeout before) and will tag a new version, so people can upgrade to PHPUnit 9.

PR #433, opened 2020-03-05
Do not overwrite existing env variables

stefandjokic:
Would like to use your repo in one of our projects.

The only thing is that setting environment variables outside runner itself during CI is important for us.

Might help others with similar setup as well.

### DIFF
--- a/phpunit.xml.dist
+++ b/phpunit.xml.dist
@@ -27,5 +27,7 @@
     <php>
         <env name="APP_ENV" value="testing"/>
         <env name="CACHE_DRIVER" value="array"/>
+        <env name="DB_CONNECTION" value="sqlite"/>
+        <env name="DB_DATABASE" value=":memory:"/>
     </php>
 </phpunit>
--- a/src/Runners/PHPUnit/BaseRunner.php
+++ b/src/Runners/PHPUnit/BaseRunner.php
@@ -196,9 +196,9 @@ abstract class BaseRunner
         $variables = $this->options->filtered['configuration']->getEnvironmentVariables();
 
         foreach ($variables as $key => $value) {
-            \putenv(\sprintf('%s=%s', $key, $value));
+            \putenv(\sprintf('%s=%s', $key, getenv($key, true) ?: $value));
 
-            $_ENV[$key] = $value;
+            $_ENV[$key] = getenv($key, true) ?: $value;
         }
     }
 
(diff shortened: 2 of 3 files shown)


stefandjokic:
@andreasschroth Any chance I can get this reviewed?

andreasschroth:
Thank you for your effort and the PR.

I'm just asking myself: Is this really expected behaviour? Most likely the environment variables are defined in the config to overwrite the existing ones? Why else would you define the environment variables inside the config?

stefandjokic:
Thank you for replying.

Well, didn't really dive into the behaviour behind the scene. We use jenkins for our CI/CD processing, during the test stage we are able to generate one time database with unique name, and pass this name as an env variable. Our project uses laravel, and we do have database name env setting in phpunit.xml for local testing purposes. But when running on jenkins, the replacement of db config values works out of box. Meaning that if there is an env variable set it will not be overwritten by phpunit.xml variables. Not sure if this has to do with how laravel loads the config or if one of those two config setups(laravel, phpunit) takes precedence, but I would say that is intended behaviour, since we did not really change anything to make it work like that. 

Here's small code snippet from our jenkins file.

`export DB_DATABASE="test_$(date +%Y%m%d%H%M%S)_$RANDOM"
export DB_USERNAME=$MYSQL_USER
export DB_PASSWORD=$MYSQL_PASSWORD
mysql -u $MYSQL_USER --password=$MYSQL_PASSWORD -e "CREATE DATABASE $DB_DATABASE;"
php -d memory_limit=512M vendor/bin/phpunit
mysql -u $MYSQL_USER --password=$MYSQL_PASSWORD -e "DROP DATABASE IF EXISTS $DB_DATABASE;"`

Thanks for the this awesome package.

stefandjokic:
Seems like it is intended behaviour not to overwrite existing env variables unless force flag is set.

Please look at https://github.com/ralphschindler/phpunit/blob/57804df84f92a19d1b6b85a2405cb71997fc69c4/src/Util/Configuration.php#L560

andreasschroth:
Thanks, I merged it.

Personally I'm not really a big fan of this change, but on the other hand it makes sense to have the same behaviour as PHPUnit here.

Thanks for your effort again.

stefandjokic:
Thank you.

Any chance I could get a release tag out of it?

PR #437, opened 2020-04-24
Add feedback item colours

slt:
Added feedback item colours similar to what phpunit does

Fixed broken test and PHPUnit 9 deprecation asserts


andreasschroth:
Thank you, looks good.

slt:
Thank you for reviewing so quickly

PR #438, opened 2020-05-02
Remove unused dependency on composer/semver

TysonAndre:
paratest stopped using it in 7689cde4460bb45162319abd4f1e48ec35c2a3a2

Previously, anything using paratest as a dev dependency would have
composer/semver unexpectedly held back to `~1.2`.
https://github.com/composer/semver/releases includes 1.5.1 and 2.0.


andreasschroth:
Thank you. 👍

PR #444, opened 2020-07-02
Fix deprecation warnings in testsuite to prepare for PhpUnit 10

Zarthus:
Replaces "assertFileNotExists" with "assertFileDoesNotExist" (1:1 compatible)
Replaces "assertRegExp" with "assertMatchesRegularExpression" (1:1 compatible)

This makes migrating to PHPUnit 10 in the future completely possible.


Zarthus:
I see now in the CI that there are multiple stages that test different PHPUnit stages (rightfully so), and that merely `composer test` would not be representative of a CI run :) so while this PR fixes the deprecation warning it probably introduces bigger problems for now and needs a more well thought-out solution, so feel free to close.

Issue #450, opened 2020-07-09
Adopt strictier version policy and stick with latest PHPUnit version only


Slamdunk:
Hi, the current README table which states that the corresponding ParaTest version to PHPUnit 9 is ParaTest 4 is bit of a lie, because the composer says `"phpunit/phpunit": "^7.5.8|^8.0|^9.0"`.

Also this forces us to test the lowest dependencies, leading to errors such https://travis-ci.org/github/paratestphp/paratest/jobs/706416438#L291 which could be easily ignored if we push users to adopt only the latest versions.

Here I propose (and would gladly make associated PR if accepted) to:

1. Support only the latest PHPUnit version with strict contraint
1. Tag a MAJOR ParaTest version for every MAJOR PHPUnit one (already done), EOLing previous ParaTest MAJOR (currently missing)
1. Tag a MINOR ParaTest version for every MINOR PHPUnit version that breaks anything in ParaTest - warnings included, EOLing previous ParaTest MINOR
1. Suggestion: for both cases, match PHPUnit and ParaTest version numbers

Benefits:

1. Drop multiple/EOL PHP version testing, we will require the same PHP version constraint as PHPUnit
1. Move the code forward quickly to match PHP and PHPUnit evolutions
1. Drop `DEPS=lowest` testing

Further suggestion:

1. Drop `PHPUNIT_DEV=true` testing: have you ever found it useful?

andreasschroth:
Thank you for your efforts and your thoughts.

Regarding your suggestions:

1. I agree, that probably makes sense.
2. Makes sense as well.
3. Yes, agreed.
4. Probably also makes sense. The only problem is, what if some changes in Paratest require a new major version, because it's a BC break?

`PHPUNIT_DEV=true`... I find it useful, but should probably be allowed to fail.

Slamdunk:
> Probably also makes sense. The only problem is, what if some changes in Paratest require a new major version, because it's a BC break?

You are right, better leave that suggestion off.

So, in #451 I've started addressing this, and dropping `DEPS=lowest` isn't feasible due to `"symfony/* "^3.4||^4.0||^5.0"`, but I've noticed that with this we are only testing `3.4` and `5.0`, but not `4.0`.
Moreover, most of this Symfony releases are already EOLed.

[...]

PR #451, opened 2020-07-12
Support latest PHPUnit only

Slamdunk:
Addresses part of #450 and should be enough for a `4.1.0` release.


andreasschroth:
Thank you, looks great. 👍

andreasschroth:
And yes, it will definitely make maintaining this a lot easier.

PR #454, opened 2020-07-17
Replace Travis-CI with Github Actions

### LINKED ISSUE #424: Switch to Github Actions

Slamdunk:
Closes #424


Slamdunk:
@andreasschroth I need you to:

1. Disable Travis CI
1. Delete Travis CI associated secrets
1. Ensure that the Settings > Actions page looks like:

immagine

Slamdunk:
Preview at https://github.com/Slamdunk/paratest/pull/1

immagine

I suppose you can't see that here for security reasons: non-owner cannot propose GA because it also have write access to repo.

Slamdunk:
Everything works as expected after merge, see https://github.com/paratestphp/paratest/pull/458.

Please just disable the Travis CI integration and delete the secret token generated for the release process, there's no need for that anymore since GA can do that without additional permissions :+1:

.github/workflows/release.yaml
```diff
@@ -0,0 +1,81 @@
...
+        run: "php box.phar compile"
+
+      - name: "Show info about paratest.phar with humbug/box"
+        run: "php box.phar info bin/paratest.phar"
+
+      - name: "Determine tag"
+        id: "determine-tag"
+        run: "echo \"::set-output name=tag::${GITHUB_REF#refs/tags/}\""
+
+      - name: "Create release"
```
Slamdunk, line 61: @andreasschroth this morning I've tested the Release Github Actions workflow.

The first time I've tried to tag the release from the Github GUI page https://github.com/paratestphp/paratest/releases/new but the GA workflow try to create that release itself and I got the error you can see at https://github.com/paratestphp/paratest/runs/917774095#step:12:9

So I did a bare `git tag` locally and pushed it, and everything worked fine with the expected PHAR attached:

1. https://github.com/paratestphp/paratest/runs/917835737
1. https://github.com/paratestphp/paratest/releases/tag/4.1.1-alpha.2

You'll see in fact the releaser is @github-actions

I don't like it very much because we can't create a release with the Github GUI anymore. The solutions may be:

1. Accept it as a new release process; if we want to add a description to the release, we'll edit the one created by GA
1. Get rid of `actions/create-release` within GA, create the release only with Github GUI so `actions/upload-release-asset` can play as expected manually providing the `upload_url`

WDYT?

Slamdunk: A third and elegant solution is to use https://github.com/laminas/automatic-releases

Everytime a milestone is closed, the release is automatically created with the list of the issue/pr closed within it.

PR #459, opened 2020-07-21
[GA] Tests: run with PCOV and SQLite

Slamdunk:
PCOV and SQLite are needed in order to run the full test suite


Slamdunk:
Before:

immagine

After:

immagine

PR #463, opened 2020-07-21
Run tests against Windows too

Slamdunk:


.github/workflows/integrate.yaml
```diff
@@ -103,6 +106,7 @@ jobs:
         run: "composer update --no-interaction --no-progress"
 
       - name: "Run tests"
+        timeout-minutes: 10
```
Slamdunk, line 109: If found many infinite loop issues on Windows, a timeout is needed

composer.json
```diff
@@ -30,8 +30,8 @@
         "phpunit/php-code-coverage": "^8.0",
         "phpunit/php-timer": "^5.0",
         "phpunit/phpunit": "^9.2",
-        "symfony/console": "^3.4 || ^4.0 || ^5.0",
-        "symfony/process": "^3.4 || ^4.0 || ^5.0"
+        "symfony/console": "^4.4 || ^5.0",
+        "symfony/process": "^4.4 || ^5.0"
```
Slamdunk, line 34: `symfony/process:3.4` simply doesn't work on Windows for this library

andreasschroth:
Thanks, looks good. 👍

PR #471, opened 2020-07-24
Add missing type hints

Slamdunk:


andreasschroth:
Thanks a lot, wanted to do this for a while already.

PR #474, opened 2020-07-27
Add missing iterable type hints

Slamdunk:


.github/workflows/integrate.yaml
```diff
@@ -275,7 +275,9 @@ jobs:
         run: "vendor/bin/phpstan analyse --no-progress --error-format=checkstyle | cs2pr"
 
       - name: "Run PSalm for type-coverage"
-        run: "vendor/bin/psalm --no-progress --shepherd --show-info=false --stats"
+        run: |
+          vendor/bin/psalm --no-progress --show-info=false > /dev/null || true
+          vendor/bin/psalm --no-progress --shepherd --show-info=false --stats
```
Slamdunk, line 280: https://github.com/vimeo/psalm/issues/3901

bin/phpunit-wrapper.php
```diff
@@ -16,18 +19,19 @@
...
-$lastExitCode = 0;
-$rand = rand ( 0 , 999999 );
-$uniqueTestToken = getenv("UNIQUE_TEST_TOKEN") ?: "no_unique_test_token";
-$testToken = getenv("TEST_TOKEN") ?: "no_test_token";
-$filename = "paratest_t-{$testToken}_ut-{$uniqueTestToken}_r-{$rand}.log";
-$path = sys_get_temp_dir()."/".$filename;
-$loggingEnabled = getenv("PT_LOGGING_ENABLE");
-$logInfo = function(string $info) use ($path, $loggingEnabled){
-    if(!$loggingEnabled){
+$commands        = fopen('php://stdin', 'r');
```
andreasschroth, line 22: I really don't like this alignment. It's not widespread and makes no sense, as it leads to unnecessary changes and diffs. E.g. when a longer variable is added here, the spaces need to be adjusted in all other lines, leading to unnecessary noise.

andreasschroth:
Thanks a lot, good changes, I just added one comment.

bin/phpunit-wrapper.php
andreasschroth: Yeah, that would be great. I really think it's pointless, as I said, one change in another line might lead to changes in the surrounding lines, which imo is bad.

PR #482, opened 2020-07-28
Remove business-logic free ParaTestApplication class

### LINKED ISSUE #455: Add compatibility with Composer v2
From https://github.com/composer/composer/blob/master/UPGRADE-2.0.md

> - `vendor/composer/installed.json` format changed:
>   - packages are now wrapped into a `"packages"` top level key instead of the whole file being the package array

This conflicts with the current method of retrieving package version:

https://github.com/paratestphp/paratest/blob/a9ca5977b3caf45e8c5956ffcd38d1e953b1421c/src/Console/VersionProvider.php#L73-L87

Slamdunk:
Closes #455 

- [x] Remove `VersionProvider`: the only usage of it is in `ParaTestApplication` constructor, but with a default version, making `VersionProvider` basically useless
- [x] Remove `ParaTestApplication`: it was introduced when `Symfony\Console` didn't provide single command application; since it contains no business logic, it can be safely replaced with a simple factory method


composer.json
```diff
@@ -21,7 +21,6 @@
     "require": {
         "php": "^7.3",
         "ext-dom": "*",
-        "ext-json": "*",
```
Slamdunk, line 24: Just love CI/CD https://github.com/paratestphp/paratest/pull/482/checks?check_run_id=918742068

PR #486, opened 2020-07-29
Escape PHP_BINARY as well

### LINKED ISSUE #483: ParaTest fails if PHP is installed on path with spaces
I'm trying to run paratest on Windows 10 where my php installation resides inside a folder with spaces in it.
This causes paratest to fail.
Here is an excerpt of the output:

    Command: D:\Program Files\PHP7\php.exe D:\Projects\XXXXXX\development\www\vendor\phpunit\phpunit\phpunit --configuration D:\Projects\XXXXXX\development\www\phpunit.xml --log-junit C:\Users\XXXXXX\AppData\Local\Temp\PT_A068.tmp D:\Projects\XXXXXX\development\www\tests\Unit\App\SomeTest.php
    StdErr: 'D:\Program' is not recognized as an internal or external command, operable program or batch file.

For now I've moved my php installation, but would be nicer if the location of my php had no influence on the working of paratest.

Slamdunk:
Closes #483

### DIFF
--- a/src/Runners/PHPUnit/Worker/BaseWorker.php
+++ b/src/Runners/PHPUnit/Worker/BaseWorker.php
@@ -11,6 +11,7 @@ use Symfony\Component\Process\PhpExecutableFinder;
 use function array_map;
 use function count;
 use function end;
+use function escapeshellarg;
 use function explode;
 use function fclose;
 use function fread;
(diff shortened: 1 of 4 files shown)


Slamdunk:
Damn Windows escaping...

PR #493, opened 2020-08-04
Pass "testName" parameter to a dataProvider method

mikolajzieba:
There is an ability in the PhpUnit to receive "testName" in the dataProvider method. This pull request enables this in paratest.


src/Runners/PHPUnit/SuiteLoader.php
```diff
@@ -263,7 +264,17 @@ private function getMethodTests(ParsedClass $class, ParsedFunction $method): arr
...
             $result            = [];
-            foreach ($testClass->$dataProvider() as $key => $value) {
+
+            $testClassReflection = new ReflectionClass($testFullClassName);
+            $dataProviderMethod  = $testClassReflection->getMethod($dataProvider);
+
+            if ($dataProviderMethod->getNumberOfParameters() === 0) {
+                $data = $dataProviderMethod->invoke($testClass);
+            } else {
+                $data = $dataProviderMethod->invoke($testClass, $method->getName());
```
Slamdunk, line 274: I guess you are referring to this: https://github.com/sebastianbergmann/phpunit/blob/9.2.6/src/Util/Annotation/DocBlock.php#L389-L393

Could you please add a test for this?

test/Unit/Runners/PHPUnit/SuiteLoaderTest.php
```diff
@@ -364,4 +364,12 @@ public function testExecutableTestsForFunctionalModeUse(): void
         $testMethod = $tests[1];
         static::assertEquals($testMethod->getName(), 'testTwoA|testTwoBDependsOnA');
     }
+
+    public function testDataProviderWithParameter(): void
+    {
+        $path   = $this->fixture('passing-tests/DataProviderParameterTest.php');
+        $loader = new SuiteLoader();
+        $loader->load($path);
+        $loader->getTestMethods();
```
Slamdunk, line 373: You should add an assertion on what you expect from this calls

mikolajzieba: @Slamdunk Yep, I'm trying to figure out your tests environment. Is this "testDataProviderWithParameter" needed anyway? As I can see, my test added in the passing-tests folder is being executed in functional tests, so maybe it should be sufficient to just leave test in passing-tests and remove my method from SuiteLoaderTest?

Slamdunk: Sounds reasonable: the code-coverage will tell you if your new lines are covered

mikolajzieba:
@Slamdunk Is there a possibility to merge it to the v3 for PHPUnit 8?

Slamdunk:
No, I'm sorry

mikolajzieba:
@Slamdunk I could prepare separate PR from 3.1.2 version, but I have no knowledge how to make a PR for a particular version upgrade.

Slamdunk:
Our policy is to maintain only the latest version of PHPUnit, because maintenance has high costs.

I suggest you to adopt a solution like https://github.com/cweagans/composer-patches to backport such fix.

mikolajzieba:
Ok, thanks for the explanation.

Issue #348, opened 2018-11-20
Bootstrap loaded before php settings are processed


bendavies:
Given a `phpunit.xml` like this:

```xml
<?xml version="1.0" encoding="UTF-8"?>

<!-- https://phpunit.de/manual/current/en/appendixes.configuration.html -->
<phpunit xmlns:xsi="http://www.w3.org/2001/XMLSchema-instance"
         xsi:noNamespaceSchemaLocation="http://schema.phpunit.de/6.5/phpunit.xsd"
         backupGlobals="false"
         colors="true"
         bootstrap="bootstrap.php"
>
    <php>
        <env name="APP_ENV" value="test" />
    </php>

    <testsuites>
        <testsuite name="Project Test Suite">
            <directory>tests</directory>
        </testsuite>
    </testsuites>
</phpunit>
```

PHPUnit will process the `<php>` block first, i.e set the `APP_ENV` env var, and then run `bootstrap.php`.

Paratest does the opposite: runs `bootstrap.php` then set the `APP_ENV` env var.

This means that if the bootstrap relies in `APP_ENV`, `phpunit` will work, while `paratest` will not.

moehrenzahn:
I can confirm this issue exists in ParaTest 2.2.0.

jccpdev:
Also seems to be broken in 2.1

Slamdunk:
Closed in #495

PR #495, opened 2020-08-04
Prove #68 is fixed

### LINKED ISSUE #68: paratest doesn't take the globals in phpunit.xml?
Hello,

For my unit tests, i need to have a specific "HTTP_HOST", so I put this in my phpunit.xml:

<php>
        <server name="HTTP_HOST" value="mywebsite.com" />
</php>

I use "$_SERVER['HTTP_HOST']" in my bootstrap.php (declared in the phpunit.xml).

It works with phpunit, but when I run it with paratest, this variable is not set.

Is this a bug or did i make something wrong?

Thank you.

Slamdunk:
Closes #68 #348

### DIFF
--- a/test/Functional/PHPUnitTest.php
+++ b/test/Functional/PHPUnitTest.php
@@ -13,9 +13,14 @@ use ParseError;
 use Symfony\Component\Console\Input\ArrayInput;
 
 use function array_key_exists;
+use function basename;
 use function chdir;
 use function dirname;
 use function file_exists;
+use function glob;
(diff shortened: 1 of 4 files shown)


hugochinchilla:
When can we expect a release containing this fix?

Slamdunk:
Well, there is no fix here, just a proof that's already fixed.

Btw, there's a lot of refactoring under the hood right now in `master`, so I'll release it only when it's ready, hopefully with as many bugfixes as possible.

PR #504, opened 2020-08-07
Allow PHPUnit 9.3

### LINKED ISSUE #503: Support for PHPUnit 9.3
Paratest cannot currently be installed alongisde PHPUnit 9.3 because the dependency on `phpunit/php-code-coverage` got bumped to `^9.0` and Paratest only supports `^8.0`.

Slamdunk:
Closes #503


Slamdunk:
I'm experiencing really slow code-coverage, see https://github.com/sebastianbergmann/phpunit/issues/4403, I'd like to wait a reply on that issue before merge

Issue #431, opened 2020-02-19
Paratest hangs indefinitely with WrapperRunner and SqliteRunner


oojacoboo:
Please see #310 as another reported case of this.  I'm unsure of why this was closed.

I've spent a little time digging further into this issue and I've identified what I think is the issue.  Paratest hangs after, what seems to be all tests, having completed.  Upon inspection, after the completion of these tests, it appears PHPUnit also has 3 processes still in the process queue with PID assignment.  The number of phpunit processes still left dangling and the number of errors reported in the paratest output are the same.

After debugging the phpunit-wrapper bin script, I've noticed that the `$lastExitCode` for the command `PHPUnit\TextUI\Command::main(false)` is `2` in all of these cases.  In these cases an Exception has been thrown.

It seems to me that, when Exceptions are being thrown and the exit status of phpunit within the `phpunit-wrapper` is not zero, that the processes aren't being handled properly.

I can manually add something like... 

```php
if ($lastExitCode != 0) {
    echo $infoText;
    exit($lastExitCode);
}
```

to the `phpunit-wrapper` within the while loop, but, this of course does not allow for the tests to complete before printing the error.  However, it's far better than having the paratest process hang indefinitely, never printing out any error details.

The funny thing is, if I alter the settings for phpunit.xml to not `stopOnError`, paratest never completes and ends up hanging much earlier in the process of tests.

Firstly, what is the preferred `phpunit.xml` config for paratest?  It doesn't seem that any configuration is entirely compatible.

And also, any insights into the process management for these `phpunit-wrapper` scripts?  If they're terminating with an exit code, how is paratest interpreting this?  It doesn't seem that the processes are exited in this case and if they are, paratest interprets this as a failure and exits itself.

Slamdunk:
I've encountered myself this behaviour many time. Yes we need to better handle wrapper/sqlite runners errors.

timglabisch:
we've the same issue. one worker seems to never stop running.

Slamdunk:
Hi @timglabisch , have you tried the latest version of ParaTest, currently 5?
If yes, can you provide a minimal reproducible environment that replicate the issue?

timglabisch:
@Slamdunk i am using ^3.1 - i'll try upgrading.

timglabisch:
@Slamdunk seems that we've the same issue with with the latest version.
Sorry but i can't provide a minimal reproducible environment. 

[...]

PR #525, opened 2020-08-28
test to proof issue 431

timglabisch:
see https://github.com/paratestphp/paratest/issues/431

### DIFF
--- a/test/Unit/Runners/PHPUnit/RunnerTestCase.php
+++ b/test/Unit/Runners/PHPUnit/RunnerTestCase.php
@@ -330,6 +330,19 @@ abstract class RunnerTestCase extends TestBase
      * @group github
      * @coversNothing
      */
+    final public function testChildProcessPipeOverflow(): void
+    {
+        $this->bareOptions['--no-test-tokens'] = true;
+        $this->bareOptions['--processes']      = 1;
+
+        $runnerResult = $this->runRunner($this->fixture('github' . DS . 'GH431'));
+        $this->assertTestsPassed($runnerResult);
+    }
+
+    /**
+     * @group github
+     * @coversNothing
+     */
     final public function testTokensArePresentByDefault(): void
     {
         $this->bareOptions['--processes'] = 1;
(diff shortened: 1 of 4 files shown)


Slamdunk:
Beside coding standards, the test are passing :\

timglabisch:
@Slamdunk could you please try to run it with `../../../../bin/paratest -c . --runner=SqliteRunner`? 
the problem just exists with the `SqliteRunner` -> is this covered by the tests?

Slamdunk:
Oh, you need to add the corresponding test to `test/Unit/Runners/PHPUnit/RunnerTestCase.php`

Slamdunk:
Locally, WrapperRunner seems ok, only SqliteRunner is failing

Slamdunk:
So, I can fix this by wrapping the test execution of sqlite-wrapper with ob_start/ob_get_clean just like the wrapper, but everything messes up again when tricking the pipes this way:

```php
        for ($i = 0; $i < 5000; ++$i) {
            fputs(\STDOUT, "Iteration: %5s\n", $i);
        }
```

timglabisch:
@Slamdunk i would prefer to read and output the pipes (and may allow to access the output). it's much easier to debug such tests when you get access to the process output. i think relying on `ob_*` is not a good idea, because in some integration tests may php's output buffer will messed up.

Slamdunk:
Symfony\Process indeed always reads the pipes:

https://github.com/symfony/process/blob/f4619b7c317b62637c57ff7cf970139a64c0bf90/Process.php#L412-L416

And of course the Runner is working because it's using `Symfony\Process`

For the records, the only reason we're using base `proc_open` in WrapperRunner/SqliteRunner is because of:

https://github.com/paratestphp/paratest/blob/b2a16958fcb9adaad29af530d532c1e7eb436a7e/src/Runners/PHPUnit/WrapperRunner.php#L115

> i think relying on `ob_*` is not a good idea, because in some integration tests may php's output buffer will messed up.

I agree

PR #527, opened 2020-09-03
Replace proc_open with Symfony\Process

### LINKED ISSUE #431: Paratest hangs indefinitely with WrapperRunner and SqliteRunner

### LINKED ISSUE #526: Replace WrapperRunner and SqliteRunner with file-based processes communication to drop proc_open in favor of Symfony\Process
Hi, WrapperRunner/SqliteRunner are the golden nugget of ParaTest: not having to kill and respawn sub-processes every test makes a huge performance difference on slow bootstrap environments.

The big issue is that they are not leveraging `Symfony\Process` component, which takes care of lot of bugs related to sub-process handling on both Linux and Windows.

SqliteRunner to me is the most well-coded, simple, easy to maintain and could easily be refactored to use `Symfony\Process`, but requires PDO and Sqlite, a big no-go to me for tools like ParaTest.

WrapperRunner suffers from depending on [`stream_select`](https://github.com/paratestphp/paratest/blob/b2a16958fcb9adaad29af530d532c1e7eb436a7e/src/Runners/PHPUnit/WrapperRunner.php#L115), which helps reducing some overheads and replace the slightly less inefficient `usleep` approach, but forbids us to use `Symfony\Process` (unless doing some nasty reflection stuff on internal attributes) and is not working on Windows.
Last but not least, pipes are a pain to handle, see https://github.com/paratestphp/paratest/issues/431

What I propose here is:

1. Drop SqliteRunner
1. Edit WrapperRunner + WrapperWorker to use `Symfony\Process`
1. Replace WrapperWorker communication with sup-process through pipes (see [`fwrite` here](https://github.com/paratestphp/paratest/blob/b2a16958fcb9adaad29af530d532c1e7eb436a7e/src/Runners/PHPUnit/Worker/WrapperWorker.php#L59) and [`fgets` here](https://github.com/paratestphp/paratest/blob/b2a16958fcb9adaad29af530d532c1e7eb436a7e/bin/phpunit-wrapper.php#L49)) with a file based one

Note that we already depend on the filesystem for logging (and code-coverage), so no new dependencies would be created.

This would (ideally) result in a WrapperRunner with a clean code (such SqliteRunner's one) and no more bugs related to sub-processes and pipes (up to ParaTest, I mean), at the cost of introducing `usleep` again as a wait system for the sub-process to end the current test, a technique already in place for Runner.
`stream_select` could in theory be adopted also for file streams, but `usleep` approach is far more easy to maintain clean in the long run.

Since I've spent the last month refactoring basically each line of the project, I'm a bit confident the proposed solution would work.
What I would like to ask you is what I cannot read from the code:

1. Has ParaTest ever tried such solution?
1. Do you see any blockers for this?

Ping:

1. Active admin: @andreasschroth 
1. Former lead developers: @brianium @julianseeger @giorgiosironi 
1. Notable contributors: @chbiel @timglabisch @michaelbutler @TysonAndre 
1. Persons which may be interested as well: @Jean85

Slamdunk:
Close #431 
Close #525 
Close #526


timglabisch:
@Slamdunk really cool, could you release a new version (tag)?

Slamdunk:
Please wait patiently, a new major version will be bumped soon, but needs a few other refactoring.

In the meanwhile you are encouraged to test this pinning `dev-master` in you composer.json :+1:

Issue #265, opened 2017-07-12
Teamcity error reporting


Strate:
Hi!

I would use this utility on teamcity server, but I can't force `--teamcity` flag to the `phpunit` runner. Is it possible?

julianseeger:
It's a little bit Note complicated.
The subprocesses of paratest report back to the main process of paratest and it is that process who needs to bei able to print teamcity result printer style. It shouldn't be complicated to add that printer to paratest. I'm just not sure if phpstorm would handle it correctly because it might not expect to see multiple Tests running at the same time.

Strate:
@julianseeger 
> I'm just not sure if phpstorm would handle it correctly because it might not expect to see multiple Tests running at the same time.

I don;t know about phpStorm, but Teamcity itself do it well with parallel tests. Nowadays I make parallel tests with GNU Parallel, and everything ok.

xakzona:
@Strate Can you give us example?

Strate:
@xakzona basically it is simple run `phpunit --teamcity $file` with help of `parallel` command

xakzona:
Paratest is better i think, the only think i need is teamcity option

andreasschroth:
Closing this now as this should be possible by using the (pretty) new `--passthru` command.

pakhomchik:
Adding `--passthru="'--teamcity'"` doesn't work as expected.

For example, `phpunit --teamcity` produces the following:
```
PHPUnit 8.4.3 by Sebastian Bergmann and contributors.
##teamcity[testCount count='42' flowId='5064']

##teamcity[testSuiteStarted name='/tests' flowId='5064']

##teamcity[testSuiteStarted name='Tests\FirstTest' locationHint='php_qn:///tests/FirstTest.php::\Tests\FirstTest' flowId='5064']

##teamcity[testStarted name='first_test' locationHint='php_qn:///tests/FirstTest.php::\Tests\FirstTest::first_test' flowId='5064']

[...]
```
But `paratest -p 3 /tests --passthru="'--teamcity'"` ignores `'--teamcity'` parameter:
```
Running phpunit in 3 processes with /vendor/phpunit/phpunit/phpunit

Configuration read from /app/phpunit.xml

E.........................................

Time: 22.12 seconds, Memory: 8.00 MB
```
`paratest -p 3 /tests --passthru="--teamcity"` returns the same result.

PR #533, opened 2020-09-26
Add --log-teamcity option

Tindarid:
In our project we need `phpunit` reporting in Teamcity format. This is not possible, because `paratest `merges test execution result and reports them in human-readable form.

However, Teamcity can manage test results by itself. So, we can use `log-teamcity` option of `phpunit` and see sorted/merged results in Teamcity.

This PR relates to this issue too: #265

### DIFF
--- a/src/Runners/PHPUnit/Options.php
+++ b/src/Runners/PHPUnit/Options.php
@@ -192,6 +192,8 @@ final class Options
     /** @var string|null */
     private $logJunit;
     /** @var string|null */
+    private $logTeamcity;
+    /** @var string|null */
     private $whitelist;
     /** @var string */
     private $tmpDir;
@@ -222,6 +224,7 @@ final class Options
         bool $functional,
         array $group,
         ?string $logJunit,
+        ?string $logTeamcity,
         ?int $maxBatchSize,
         bool $noTestTokens,
         bool $parallelSuite,
@@ -254,6 +257,7 @@ final class Options
         $this->functional        = $functional;
         $this->group             = $group;
         $this->logJunit          = $logJunit;
+        $this->logTeamcity       = $logTeamcity;
         $this->maxBatchSize      = $maxBatchSize;
         $this->noTestTokens      = $noTestTokens;
         $this->parallelSuite     = $parallelSuite;
@@ -365,6 +369,10 @@ final class Options
             if ($options['log-junit'] === null && $logging->hasJunit()) {
                 $options['log-junit'] = $logging->junit()->target()->path();
             }
+
+            if ($options['log-teamcity'] === null && $logging->hasTeamCity()) {
+                $options['log-teamcity'] = $logging->teamCity()->target()->path();
+            }
         }
 
         if ($configuration !== null) {
@@ -395,6 +403,7 @@ final class Options
             $options['functional'],
             $group,
             $options['log-junit'],
+            $options['log-teamcity'],
             (int) $options['max-batch-size'],
             $options['no-test-tokens'],
             $options['parallel-suite'],
@@ -529,6 +538,12 @@ final class Options
                 null,
                 InputOption::VALUE_REQUIRED,
                 'Log test execution in JUnit XML format to file.'
+            ),
+            new InputOption(
+                'log-teamcity',
+                null,
+                InputOption::VALUE_REQUIRED,
+                'Log test execution in Teamcity format to file.'
             ),
             new InputOption(
                 'max-batch-size',
@@ -885,6 +900,16 @@ final class Options
     public function logJunit(): ?string
     {
         return $this->logJunit;
+    }
+
+    public function logTeamcity(): ?string
+    {
+        return $this->logTeamcity;
+    }
+
+    public function hasLogTeamcity(): bool
+    {
+        return $this->logTeamcity !== null;
     }
 
     public function tmpDir(): string
--- a/src/Runners/PHPUnit/ResultPrinter.php
+++ b/src/Runners/PHPUnit/ResultPrinter.php
@@ -15,7 +15,11 @@ use function array_filter;
 use function array_map;
 use function assert;
 use function count;
+use function fclose;
+use function file_get_contents;
 use function floor;
+use function fopen;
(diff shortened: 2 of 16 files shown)


Tindarid:
IDK why coverage report fails - you can see that, for example, `teamcity` function is used

And i am discouraged by message about `LineCoverage` from static analysis step

src/Runners/PHPUnit/Options.php
```diff
@@ -820,6 +831,11 @@ public function parallelSuite(): bool
         return $this->parallelSuite;
     }
 
+    public function teamcity(): bool
+    {
+        return $this->teamcity;
```
Slamdunk, line 836: This needs to be tested in `\ParaTest\Tests\Unit\Runners\PHPUnit\OptionsTest`

src/Runners/PHPUnit/WrapperRunner.php
```diff
@@ -24,6 +24,12 @@ final class WrapperRunner extends BaseRunner
 
     protected function beforeLoadChecks(): void
     {
+        if ($this->options->teamcity()) {
+            throw new InvalidArgumentException(
+                'The `teamcity` option is not supported yet in the WrapperRunner.'
```
Slamdunk, line 29: May I ask you why?

src/Runners/PHPUnit/ResultPrinter.php
```diff
@@ -201,6 +202,22 @@ public function printFeedback(ExecutableTest $test): Reader
         return $reader;
     }
 
+    /**
+     * Prints the individual feedback in teamcity form
+     */
+    public function printTeamcityFeedback(ExecutableTest $test): void
```
Slamdunk, line 208: This method must also be tested within `\ParaTest\Tests\Unit\Runners\PHPUnit\ResultPrinterTest`.

Slamdunk (changes requested):
I've pushed in master a fix for Static Analysis [feedbacks](https://github.com/paratestphp/paratest/pull/534), please rebase.

Slamdunk:
@Tindarid thanks for the rebase; did you notice that my requested changes contain also other 3 comments?

Tindarid:
Yep, I noticed. I am thinking about exit code of paratest depending on exit code of phpunit. If we do not manage results and leave this to teamcity - paratest will always exit with 0. Is it a problem?

Also, maybe it will be a good decision to also log to junit and report exit code correctly

Slamdunk:
Indeed I was thinking that it would be better to keep the default behavior of xml-log and standard-output, drop the proposed `--teamcity` options and only add a `--log-teamcity` option.

In your case you'll have to read the file instead of direct paratest output.

Can you give it a try?

Tindarid:
I think `mkfifo FILE` + `--log-teamcity FILE` will be great too

So, phpunit supports passing `--log-teamcity` and `--log-junit` at the same time

And in case of `--log-teamcity FILE` we will pass `--log-teamcity` to phpunit (and `--log-junit`) and merge results in `FILE` (everything else will not change)
And there can be `--teamcity` option, where paratest uses `--log-junit` result to have a line in the end, like `Tests: OK (...)` and merge results of '--log-teamcity` in stdout (maybe in some other PR)

Is it OK?

Slamdunk:
> I think `mkfifo FILE` + `--log-teamcity FILE` will be great too
> 
> So, phpunit supports passing `--log-teamcity` and `--log-junit` at the same time
> 
> And in case of `--log-teamcity FILE` we will pass `--log-teamcity` to phpunit (and `--log-junit`) and merge results in `FILE` (everything else will not change)

Exactly

> And there can be `--teamcity` option, where paratest uses `--log-junit` result to have a line in the end, like `Tests: OK (...)` and merge results of '--log-teamcity` in stdout (maybe in some other PR)

I'd rather not to touch STDOUT at all, the code+tests can quickly became messy.
Teamcity is only useful to machines, so whenever you need it, a simple `vendor/bin/paratest --log-teamcity /tmp/file && cat /tmp/file` is enough, so no `--teamcity` option please.

> Is it OK?

Yup, thanks for working on this.

src/Runners/PHPUnit/Options.php
Tindarid: not relevant, tests for `--log-teamcity` exists

src/Runners/PHPUnit/ResultPrinter.php
Tindarid: not relevant, tests for `--log-teamcity` exists

src/Runners/PHPUnit/WrapperRunner.php
Tindarid: Now, `--log-teamcity` option supports it

Slamdunk:
Please next time do not squash commits

Tindarid:
Sorry for that, I haven't managed to set up environment to full test it locally (eg: covertage report)

Slamdunk:
Still, please stop squashing your commits, it messes every review up

Slamdunk:
Thank you @Tindarid 

Had to open another PR because you commited on your fork master branch and that doesn't allow me further rebase nor commits that were needed. Note that I need https://github.com/sebastianbergmann/phpunit/pull/4446 to get merged before release a new MAJOR, but you can start using this tagging `dev-master` on your composer.json.

Tindarid:
Thank you

Tindarid:
Sorry, but your changes with filesystem functions break my view of functionality.

When you call `put_contents` it reopens file, but i want to make FIFO pipe and write to it. So on call `put_contents` it places EOF to the end, and it breaks everything

Could you revert this changes, please?

Slamdunk:
> Could you revert this changes, please?

I haven't change the tests: if I revert the commits, the bug may reappear unnoticed.

Please provide a complete expected-vs-actual result, so we can write a dedicated test case for the bug and find the best solution to it.

Slamdunk:
Ping @Tindarid

Tindarid:
> Ping @Tindarid

https://github.com/paratestphp/paratest/pull/544

Sorry, for waiting

PR #539, opened 2020-10-14
Add paratest tests execution in random sort

proggga:
Added --order-by by default|random|reversed
If random selected - will be chosen random seed and printed to user (to reproduce)
--random-order-seed=123 can be used to pass custom seed for reproducing
also added reversed for reversed test passing 

P.S. this parameter doesn't passed right to phpunit, it applies only for paratest, if someone want use phpunit test random he can pass --order-by in passthru param

FYI
removed `assert($executebleTest !== null);` code from Runner.php cause Static Analysis (7.4) Failed, and failed at master branch too, so I removed this assert (you can see error msg [here](https://github.com/paratestphp/paratest/runs/1254349339?check_suite_focus=true))
If I wrong please help me with one

PR #540, opened 2020-10-14
Fix phpunit-wrapper from resetting variables in global scope in WrapperRunner

proggga:
phpunit-wrapper has a problem

it's broken in WrapperRunner
phpunit reset global variables between runs
https://phpunit.readthedocs.io/en/9.3/fixtures.html#global-state

I've got problems with global state in phpunit and this fix helped me.
I tried to user WrapperRunner in my big project, and this helps

this fix prevent variables in phpwrapper (they all in global scope) from turning to null 

(Also I removed assert here like in #539, I need this for passing all tests)

### DIFF
--- a/bin/phpunit-wrapper.php
+++ b/bin/phpunit-wrapper.php
@@ -2,39 +2,41 @@
 
 declare(strict_types=1);
 
-$opts = getopt('', [
-    'write-to:',
-]);
+function phpunitWrapperMain(): void {
 
-$composerAutoloadFiles = [
+  $opts = getopt('', [
+    'write-to:',
+  ]);
+
+  $composerAutoloadFiles = [
     dirname(__DIR__, 3) . DIRECTORY_SEPARATOR . 'autoload.php',
     dirname(__DIR__, 2) . DIRECTORY_SEPARATOR . 'vendor' . DIRECTORY_SEPARATOR . 'autoload.php',
     dirname(__DIR__) . DIRECTORY_SEPARATOR . 'vendor' . DIRECTORY_SEPARATOR . 'autoload.php',
-];
+  ];
 
-foreach ($composerAutoloadFiles as $file) {
+  foreach ($composerAutoloadFiles as $file) {
     if (file_exists($file)) {
-        require_once $file;
-        define('PHPUNIT_COMPOSER_INSTALL', $file);
+      require_once $file;
+      define('PHPUNIT_COMPOSER_INSTALL', $file);
 
-        break;
+      break;
     }
-}
+  }
 
-assert(is_string($opts['write-to']));
-$writeTo = fopen($opts['write-to'], 'wb');
-assert(is_resource($writeTo));
+  assert(is_string($opts['write-to']));
+  $writeTo = fopen($opts['write-to'], 'wb');
+  assert(is_resource($writeTo));
 
-$i = 0;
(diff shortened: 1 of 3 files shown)


src/Runners/PHPUnit/Runner.php
```diff
@@ -65,8 +65,6 @@ private function fillRunQueue(array $availableTokens): void
             && ($token = array_shift($availableTokens)) !== null
         ) {
             $executebleTest = array_shift($this->pending);
-            /** @psalm-suppress RedundantConditionGivenDocblockType **/
-            assert($executebleTest !== null);
```
Slamdunk, line 69: If you leave this here, what test is failing?

Slamdunk (changes requested):
I get the idea of the issue, but please can you try to write a test for the bug?
It seems an important one, I'd like to prevent it from being reintroduced by a future refactor

src/Runners/PHPUnit/Runner.php
proggga: Please look at  #539, I wrote there more info about problem

proggga:
> I get the idea of the issue, but please can you try to write a test for the bug?
> It seems an important one, I'd like to prevent it from being reintroduced by a future refactor

It's problematic, cause phpunit-wrapper fails, and paratest doing nothing, he is waiting for some command or idk what. It will never complete.

src/Runners/PHPUnit/Runner.php
Slamdunk: Thank you for bringing this into my attention.
I've pushed to master the change, so now you cane rebase the PR onto it

Slamdunk:
> It's problematic, cause phpunit-wrapper fails, and paratest doing nothing, he is waiting for some command or idk what. It will never complete.

That's ok: the pipeline has `timeout-minutes: 3` set, so it will fail is a test is hanging forever.

Please try to write a test so I can see the pipeline failing

proggga:
> > It's problematic, cause phpunit-wrapper fails, and paratest doing nothing, he is waiting for some command or idk what. It will never complete.
> 
> That's ok: the pipeline has `timeout-minutes: 3` set, so it will fail if a test is hanging forever.
> 
> Please try to write a test so I can see the pipeline failing

Okey, I written test, that fails on master branch (branch without my fix)
problem was in param in phpunit "backupGlobal=true", you can check this out

composer.json
```diff
@@ -37,14 +37,14 @@
         "ekino/phpstan-banned-code": "^0.3.1",
         "ergebnis/phpstan-rules": "^0.15.2",
         "infection/infection": "^0.17.3",
-        "phpstan/phpstan": "^0.12.45",
+        "phpstan/phpstan": "^0.12.49",
```
Slamdunk, line 40: The [master branch in your fork](https://github.com/proggga/paratest/tree/master) is way outdated: you should pull the latest commits from upstream, and then rebase the branch onto it, so your PRs don't get polluted by this kind of changes

proggga:
Also pls add label after all "hacktoberfest-accepted"

composer.json
proggga: fixed

Slamdunk (approved):

Slamdunk:
Thank you @proggga

PR #543, opened 2020-10-15
Add Contributors file

proggga:


Slamdunk:
Thanks but I don't find such kind of file useful.

proggga:
I think we should praise contributors. It will show this repo more friendly for contributors.

PR #544, opened 2020-10-18
--log-teamcity option fix and test

Tindarid:
Refers to https://github.com/paratestphp/paratest/pull/533

Paratest will not work as expected, if teamcity file is FIFO file.

The test is too bad for merging, but the fix shows that test will pass. Test is bad, because of lack of good FIFO features in php (no non-blocking fopen on Windows for example). But maybe we can adapt it.

### DIFF
--- a/src/Runners/PHPUnit/ResultPrinter.php
+++ b/src/Runners/PHPUnit/ResultPrinter.php
@@ -112,13 +112,21 @@ final class ResultPrinter
     private $output;
     /** @var Options */
     private $options;
+  /**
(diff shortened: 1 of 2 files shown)


Slamdunk:
Things should be ok in `master` now, can you check it please?

Tindarid:
> Things should be ok in `master` now, can you check it please?

Yep, all is excellent. `mkfifo` + `paratest` + `cat` works as expected. Thank you!

Could you add `hacktoberfest-accepted` to something related to my PRs?

PR #548, opened 2020-10-22
Add support of env variable RANDOM_ORDER_SEED

proggga:
I really need to pass Env variable RANDOM_ORDER_SEED to paratest and to test cases.
This allow me to pass by env variable RANDOM_ORDER_SEED to paratest and phpunit.
By this fix, I will mock random functions , and they will be depend on randomseed

for example
export RANDOM_ORDER_SEED=$(date +%s) 
./vendor/bin/paratest --order-by=random

in testcase

function test_randomTest() {
   $this->mock('randomChanse')->willReturnCallback(func() {
       mt_srand(getenv('RANDOM_ORDER_SEED') ?: 123);
       return mt_rand(1, 100500);
 }

availability to use the same seed for test and shuffle give possibility to reproduce any random tests

also it very useful to pass int from env, not from param

### DIFF
--- a/phpstan.neon.dist
+++ b/phpstan.neon.dist
@@ -112,18 +112,21 @@ parameters:
             count: 1
             path: test/fixtures/wrapper_runner_exit_code_tests/FailureTest.php
         -
-            message: "#^Offset 'TEST_TOKEN' does not exist on array\\('PARATEST' \\=\\> 1, \\?'TEST_TOKEN' \\=\\> int, \\?'UNIQUE_TEST_TOKEN' \\=\\> string\\)\\.$#"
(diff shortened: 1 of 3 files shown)


Slamdunk:
Hi, I'm going to reject this PR for 2 reasons:

1. Randomness is already deterministic with `--random-order-seed` set, as proven by https://github.com/paratestphp/paratest/pull/549/files#diff-0b8282eab1a963b31340842fbdb2e92db3f4216ad5ece8ba60e49c826c48b10fR23: you should simply never call `mt_srand` by yourself
1. We only support features that PHPUnit supports; maintaining this tool is hard and time consuming, no specific feature is accepted

PR #570, opened 2020-12-07
Output skipped details on verbose output

Slamdunk:


src/Runners/PHPUnit/Options.php
```diff
@@ -705,10 +720,9 @@ public static function setInputDefinition(InputDefinition $inputDefinition): voi
             ),
             new InputOption(
                 'verbose',
-                'v',
-                InputOption::VALUE_REQUIRED,
-                'If given, debug output is printed. Example: --verbose=1',
-                0
+                'v|vv',
+                InputOption::VALUE_NONE,
```
veewee, line 724: FYI : This introduces a breaking change for the end user:

```
   [Symfony\Component\Console\Exception\RuntimeException]  
  The "--verbose" option does not accept a value.         

```

No worries, I'll fix it in my scripts - but wanted to share this for people with similar issues.

Slamdunk: You are right. As this is a tool and not a library, and the flag bugged as it never showed the skipping tests like phpunit, I didn't bother much, I'm honest

PR #575, opened 2021-01-14
Prepare for PHPUnit 10

Slamdunk:


composer.json
```diff
@@ -24,10 +24,10 @@
         "ext-pcre": "*",
         "ext-reflection": "*",
         "ext-simplexml": "*",
-        "phpunit/php-code-coverage": "^9.2.5",
+        "phpunit/php-code-coverage": "10.0.x-dev",
         "phpunit/php-file-iterator": "^3.0.5",
         "phpunit/php-timer": "^5.0.3",
-        "phpunit/phpunit": "^9.5.0",
+        "phpunit/phpunit": "10.0.x-dev",
```
Slamdunk, line 30: Update once PR ready

composer.json
```diff
@@ -42,11 +42,17 @@
         "phpstan/phpstan-deprecation-rules": "^0.12.6",
         "phpstan/phpstan-phpunit": "^0.12.17",
         "phpstan/phpstan-strict-rules": "^0.12.7",
+        "sebastian/code-unit": "2.0.x-dev",
```
Slamdunk, line 45: Remove once PR ready

nunomaduro:
@Slamdunk For info, PHPUnit 10 is set to be released in early February. Wondering if you have any plans to move forward with this pull request?

Slamdunk:
Huh, didn't know that, I'll get here soon then 🙂

Slamdunk:
Just so you know, that's a major blocker currently: https://github.com/sebastianbergmann/phpunit/issues/4993#issuecomment-1378838105

oprypkhantc:
@Slamdunk 

Hey. I'm currently in the process of implementing a custom runner for Paratest, one that is similar to `WrapperRunner` but only spawns one `phpunit` process that runs multiple tests.

The reason I'm doing that is because we have a huge monolith projects with 10k+ long running tests and a long "cold startup" application time. This means that not only does running `phpunit` take a long time because of the time it takes to scan the whole codebase for tests, it also takes a lot of time for every test executed as it requires an init of our whole application.

While doing that, I saw that PHPUnit 10 was about to come out and it includes an event system. That would fit us perfectly; we'd be able to init the application once (in "test suite started" event) and then just use it in every test, without having to init it over and over again 10k+ times.

I'm aware there are limitations to this that `phpunit` sets - namely memory leaks in 9.x, but those aren't that big and will eventually be solved. Worse case scenario, I could just limit the max amount of tests per `phpunit` run or increase memory limit.

So I'm proposing the following: instead of running the whole PHPUnit (using `(new PHPUnit\TextUI\Command())->run($arguments, false);` in `wrapper-runner`) per every test executed, why not split tests ahead of time and run PHPUnit once? That would solve my problem, make `paratest` much more performant for cases like ours and most importantly - does not require any internal PHPUnit APIs; running a `phpunit` process directly will be more than enough.

What are you thinking? If this is something you're interested in, I'll be more than happy to contribute.

Slamdunk:
> why not split tests ahead of time and run PHPUnit once?

The goal of ParaTest is to execute the whole test suite as quick as possible, spreading tests among multiple processes.
How do you know the execution time of each tests, before running them?
If you don't know how much each test takes, you cannot calculate how to split them preemptively. You would end up with a process that already finished, and another one that still has to execute dozens of tests.

If you already know the best way to split your test suite into chunks that take the same amount of time, you don't need ParaTest, you only need `parallel` linux built-in binary.

oprypkhantc:
> > why not split tests ahead of time and run PHPUnit once?
> 
> The goal of ParaTest is to execute the whole test suite as quick as possible, spreading tests among multiple processes. How do you know the execution time of each tests, before running them? If you don't know how much each test takes, you cannot calculate how to split them preemptively. You would end up with a process that already finished, and another one that still has to execute dozens of tests.
> 
> If you already know the best way to split your test suite into chunks that take the same amount of time, you don't need ParaTest, you only need `parallel` linux built-in binary.

It makes sense, we don't know the time for each test either. However, our test suite takes roughly 30-40 minutes to complete in 8 threads on CI, so even if one of the processes takes longer to finish than others it would still be beneficial to us. We can't use simple `parallel` because Paratest does a great job at combining test results, exit codes and most importantly - code coverage.

But I got your point, this won't fit most use cases of Paratest. I think with the way WrapperRunner has `phpunit-wrapper` running the whole time, I can use workarounds to achieve what I want - namely I can just hold the application instance in a static variable and that should solve it.

Thank you!

rcerljenko:
hi guys... it's out...  :)

https://phpunit.de/announcements/phpunit-10.html

PR #584, opened 2021-01-29
Add support for no-coverage option

bastien-phi:
This PR makes possible to use a `--no-coverage` option which make paratest ignore coverage configuration.

It is useful to ignore coverage configuration defined in xml configuration, for example if there are no coverage driver available.

### DIFF
--- a/src/Runners/PHPUnit/Options.php
+++ b/src/Runners/PHPUnit/Options.php
@@ -1045,11 +1045,6 @@ final class Options
         return $this->coverageXml;
     }
 
-    public function noCoverage(): bool
-    {
-        return $this->noCoverage;
-    }
-
     public function cwd(): string
     {
         return $this->cwd;
--- a/src/Runners/PHPUnit/Options.php
+++ b/src/Runners/PHPUnit/Options.php
@@ -203,6 +203,8 @@ final class Options
     private $coverageText;
     /** @var string|null */
     private $coverageXml;
+    /** @var bool */
+    private $noCoverage;
     /** @var string */
     private $cwd;
     /** @var string|null */
@@ -247,6 +249,7 @@ final class Options
         ?string $logJunit,
         ?string $logTeamcity,
         ?int $maxBatchSize,
+        bool $noCoverage,
         bool $noTestTokens,
         bool $parallelSuite,
         ?array $passthru,
@@ -283,6 +286,7 @@ final class Options
         $this->logJunit          = $logJunit;
         $this->logTeamcity       = $logTeamcity;
         $this->maxBatchSize      = $maxBatchSize;
+        $this->noCoverage        = $noCoverage;
         $this->noTestTokens      = $noTestTokens;
         $this->parallelSuite     = $parallelSuite;
         $this->passthru          = $passthru;
@@ -319,6 +323,7 @@ final class Options
         assert(is_bool($options['functional']));
         assert($options['log-junit'] === null || is_string($options['log-junit']));
         assert($options['log-teamcity'] === null || is_string($options['log-teamcity']));
+        assert(is_bool($options['no-coverage']));
         assert(is_bool($options['no-test-tokens']));
         assert($options['order-by'] === null || is_string($options['order-by']));
         assert(is_bool($options['parallel-suite']));
@@ -516,6 +521,7 @@ final class Options
             $options['log-junit'],
             $options['log-teamcity'],
             (int) $options['max-batch-size'],
+            $options['no-coverage'],
             $options['no-test-tokens'],
             $options['parallel-suite'],
             self::parsePassthru($options['passthru']),
@@ -536,6 +542,10 @@ final class Options
 
     public function hasCoverage(): bool
     {
+        if ($this->noCoverage) {
+            return false;
+        }
+
         return $this->coverageClover !== null
             || $this->coverageCobertura !== null
             || $this->coverageCrap4j !== null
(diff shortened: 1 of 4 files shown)


bastien-phi:
Not sure why static analysis is failing, here is the report with affected files and lines 
```

 ------ ----------------------------------------------------------------------------------------------------- 
  Line   src/Logging/JUnit/Reader.php                                                                         
 ------ ----------------------------------------------------------------------------------------------------- 
  186    Call to function assert() with true will always evaluate to true.                                    
  186    Strict comparison using !== between array<SimpleXMLElement> and false will always evaluate to true.  
  208    Call to function assert() with true will always evaluate to true.                                    
  208    Strict comparison using !== between array<SimpleXMLElement> and false will always evaluate to true.  
  212    Call to function assert() with true will always evaluate to true.                                    
  212    Strict comparison using !== between array<SimpleXMLElement> and false will always evaluate to true.  
 ------ ----------------------------------------------------------------------------------------------------- 

 ------ ---------------------------------------------------------------------------------------------- 
  Line   src/Logging/JUnit/TestCase.php                                                                
 ------ ---------------------------------------------------------------------------------------------- 
  96     Casting to array<SimpleXMLElement> something that's already array<SimpleXMLElement>.          
  115    Casting to array<SimpleXMLElement> something that's already array<SimpleXMLElement>.          
  117    Casting to array<SimpleXMLElement> something that's already array<SimpleXMLElement>.          
  118    Casting to array<SimpleXMLElement> something that's already array<SimpleXMLElement>.          
  143    Call to function assert() with true will always evaluate to true.                             
  143    Strict comparison using !== between SimpleXMLElement and false will always evaluate to true.  
 ------ ---------------------------------------------------------------------------------------------- 

 ------ ----------------------------------------------------------------------------------------------------------------------------- 
  Line   test/Unit/Runners/PHPUnit/BaseRunnerTest.php                                                                                 
 ------ ----------------------------------------------------------------------------------------------------------------------------- 
  131    Call to static method PHPUnit\Framework\Assert::assertNotFalse() with array<SimpleXMLElement> will always evaluate to true.  
  133    Call to static method PHPUnit\Framework\Assert::assertNotFalse() with array<SimpleXMLElement> will always evaluate to true.  
  135    Call to static method PHPUnit\Framework\Assert::assertNotFalse() with array<SimpleXMLElement> will always evaluate to true.  
  137    Call to static method PHPUnit\Framework\Assert::assertNotFalse() with array<SimpleXMLElement> will always evaluate to true.  
  139    Call to static method PHPUnit\Framework\Assert::assertNotFalse() with array<SimpleXMLElement> will always evaluate to true.  
  141    Call to static method PHPUnit\Framework\Assert::assertNotFalse() with array<SimpleXMLElement> will always evaluate to true.  
 ------ -----------------------------------------------------------------------------------------------------------------------------

File: test/Unit/Runners/PHPUnit/OptionsTest.php [hunk 1]
```diff
@@ -346,7 +349,7 @@ public function testProvidedOptions(): void
         ], $options->filtered());
 
         static::assertTrue($options->hasLogTeamcity());
-        static::assertTrue($options->hasCoverage());
+        static::assertFalse($options->hasCoverage());
```
Slamdunk, line 352: If you change this, there are no place where we test `static::assertTrue($options->hasCoverage());`, can you rearrange test to have both cases please?

Slamdunk (changes requested):
I've fixed in upstream the PHPStan errors, please rebase

test/Unit/Runners/PHPUnit/OptionsTest.php
bastien-phi: On it !

bastien-phi: I just saw that it is tested in `testGatherOptionsFromConfiguration` https://github.com/paratestphp/paratest/pull/584/files#diff-ed06b3688f45d43818a780f94b65eb6bfdb6781cbe77f15d2135cfd4c4b146a6R387

Do I still need to add another test when coverage is enabled from command lines options ?

src/Runners/PHPUnit/Options.php
```diff
@@ -1029,6 +1045,11 @@ public function coverageXml(): ?string
         return $this->coverageXml;
     }
 
+    public function noCoverage(): bool
```
Slamdunk, line 1048: On a second analisys, this is not needed, since we only use `hasCoverage` method

test/Unit/Runners/PHPUnit/OptionsTest.php
Slamdunk: Yes: `testGatherOptionsFromConfiguration` tests, as it says, options from configuration.

We need instead both `assertTrue` and `assertFalse` over `$options->hasCoverage()` in the pure CLI options scenario.

So please revert this one, and create a separate test for the new case

Slamdunk (changes requested):

src/Runners/PHPUnit/Options.php
bastien-phi: Solved by 4067542

test/Unit/Runners/PHPUnit/OptionsTest.php
bastien-phi: Solved by b6cd92f

bastien-phi:
Not sure why coverage failed... Anything I did wrong ?

Slamdunk:
Cache, cache issues everywhere... Thanks for PR

bastien-phi:
Thanks for accepting it !

ibrunotome:
I was just looking for that option! Any prediction of when it will be tagged?

Slamdunk:
Released now in [v6.2.0](https://github.com/paratestphp/paratest/releases/tag/v6.2.0)

PR #589, opened 2021-02-10
Use a wrapper test suite to avoid TestSuiteLoader problems

slt:
PHPUnit's TestSuiteLoader will expect to be loading the
test class for the very first time, as Paratest uses
runs multiple phpunit commands within the same php
process, classes may already be loaded.

This happens with tests that reference other test classes
i.e. constants, static methods, inheritance etc.
causing paratest to crash :(

By dynamically generating a testsuite class for each
class we want to execute, we can avoid these problems
giving PHPunit a unique class to load each time.

### DIFF
--- a/README.md
+++ b/README.md
@@ -24,7 +24,7 @@ Why use `paratest` over the alternative parallel test runners out there?
 To install with composer run the following command:
 
     composer require --dev brianium/paratest
-    
+
 # Versions
 
 Only the latest version of PHPUnit is supported, and thus only the latest version of ParaTest is actively maintained.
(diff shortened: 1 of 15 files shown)


Slamdunk:
Hi, thank for the effort you put in this PR.
I am nevertheless incline to reject this for two reasons:

1. The underlying bug is not trivial, but trivial is the solution, which also cleans a lot the code base: put shared code into classes which are not tests _themselves_, as you know since you've deleted the `Caveats` doc
2. PHPUnit 10 should by the way fix this natively, see https://github.com/sebastianbergmann/phpunit/pull/4583

slt:
Thanks, I am very pleased to see this getting fixed in the core of phpunit - I had the impression  from https://github.com/sebastianbergmann/phpunit/pull/4037 that this was not going to be fixed, which is why I've attempted to do it here. 

Cleaning up and enforcing that test cases cannot reference each other sounds trivial, but it would have been mammoth task in my project unfortunately.

PR #594, opened 2021-02-17
Implemented `repeat` option

stephandesouza:
Allow Paratest to run multiple times, as `--repeat` present on PHPUnit

image

I could not find a better way to:

- [ ] avoid incremental test count
- [ ] avoid the `Single-run is self-completed` part

### DIFF
--- a/phpstan-baseline.neon
+++ b/phpstan-baseline.neon
@@ -96,7 +96,7 @@ parameters:
 
 		-
 			message: "#^Cannot cast array\\<string\\>\\|bool\\|int\\|string\\|null to int\\.$#"
-			count: 3
+			count: 4
 			path: src/Runners/PHPUnit/Options.php
 
 		-
--- a/phpstan-baseline.neon
+++ b/phpstan-baseline.neon
@@ -95,13 +95,8 @@ parameters:
 			path: src/Runners/PHPUnit/Options.php
 
 		-
-			message: "#^Parameter \\#37 \\$repeat of class ParaTest\\\\Runners\\\\PHPUnit\\\\Options constructor expects string\\|null, array\\<string\\>\\|bool\\|int\\|string\\|null given\\.$#"
-			count: 1
-			path: src/Runners/PHPUnit/Options.php
-
-		-
 			message: "#^Cannot cast array\\<string\\>\\|bool\\|int\\|string\\|null to int\\.$#"
-			count: 3
+			count: 4
 			path: src/Runners/PHPUnit/Options.php
 
 		-
--- a/phpstan-baseline.neon
+++ b/phpstan-baseline.neon
@@ -95,8 +95,13 @@ parameters:
 			path: src/Runners/PHPUnit/Options.php
 
 		-
+			message: "#^Parameter \\#37 \\$repeat of class ParaTest\\\\Runners\\\\PHPUnit\\\\Options constructor expects string\\|null, array\\<string\\>\\|bool\\|int\\|string\\|null given\\.$#"
+			count: 1
+			path: src/Runners/PHPUnit/Options.php
+
+		-
 			message: "#^Cannot cast array\\<string\\>\\|bool\\|int\\|string\\|null to int\\.$#"
-			count: 4
+			count: 3
 			path: src/Runners/PHPUnit/Options.php
 
 		-
--- a/src/Runners/PHPUnit/Options.php
+++ b/src/Runners/PHPUnit/Options.php
@@ -219,7 +219,7 @@ final class Options
     private $orderBy;
     /** @var int */
     private $randomOrderSeed;
-    /** @var string|null */
+    /** @var int */
     private $repeat;
 
     /**
@@ -267,7 +267,7 @@ final class Options
         ?string $whitelist,
         string $orderBy,
         int $randomOrderSeed,
-        ?string $repeat
+        int $repeat
     ) {
         $this->bootstrap         = $bootstrap;
         $this->colors            = $colors;
@@ -431,9 +431,9 @@ final class Options
         }
 
         if (is_string($options['repeat'])) {
-            if (! is_numeric($options['repeat'])) {
+            if ($options['repeat'] !== (string) (int) $options['repeat']) {
                 throw new InvalidArgumentException(sprintf(
-                    'Option --repeat should have a number value, "%s" given',
+                    'Option --repeat should have an integer value, "%s" given',
                     $options['repeat']
                 ));
             }
@@ -553,7 +553,7 @@ final class Options
             $options['whitelist'],
             $options['order-by'] ?? self::ORDER_DEFAULT,
             (int) $options['random-order-seed'],
-            $options['repeat']
+            (int) $options['repeat']
         );
     }
 
@@ -758,7 +758,7 @@ final class Options
             ),
             new InputOption(
                 'repeat',
-                'r',
+                null,
                 InputOption::VALUE_OPTIONAL,
                 'Runs the test(s) repeatedly.'
             ),
@@ -1108,7 +1108,7 @@ final class Options
         return $this->randomOrderSeed;
     }
 
-    public function repeat(): ?string
+    public function repeat(): int
     {
         return $this->repeat;
     }
--- a/src/Runners/PHPUnit/Options.php
+++ b/src/Runners/PHPUnit/Options.php
@@ -219,7 +219,7 @@ final class Options
     private $orderBy;
     /** @var int */
     private $randomOrderSeed;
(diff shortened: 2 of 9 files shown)


Slamdunk:
Instead of running the suite 3 times, what about duplicating the suites/files *n* times, here https://github.com/paratestphp/paratest/blob/2fd895d4e255a612a2d323185e93f46d6cef6fb5/src/Runners/PHPUnit/SuiteLoader.php#L168-L172 ?

stephandesouza:
> Instead of running the suite 3 times, what about duplicating the suites/files _n_ times, here
> 
> https://github.com/paratestphp/paratest/blob/2fd895d4e255a612a2d323185e93f46d6cef6fb5/src/Runners/PHPUnit/SuiteLoader.php#L168-L172
> 
> ?

I need to apply some tests, but I guess this will not work when filtering the tests.

Slamdunk:
>  I guess this will not work when filtering the tests.

Why not?

stephandesouza:
> > I guess this will not work when filtering the tests.
> 
> Why not?

Suites are indexed by its filename/suitename on `initSuites()`, I guess, to prevent miss duplications.

Running with 1 (default) repeat, focus on config.
image

Running part with 3 repeat, focus on proposed change
image

Slamdunk:
And what about simply `--passthru` it like other options like `--group`:

https://github.com/paratestphp/paratest/blob/2fd895d4e255a612a2d323185e93f46d6cef6fb5/src/Runners/PHPUnit/Options.php#L417

```
$ vendor/bin/paratest
Running phpunit in 4 processes with /var/www/html/vendor/phpunit/phpunit/phpunit                                                                                                                                                                
                                                                                                                                                                                                                                                
Configuration read from /var/www/html/phpunit.xml                                                                                                                                                                                               
                                                                                                                                                                                                                                                
....................                                              20 / 20 (100%)

Time: 00:09.778, Memory: 16.00 MB

OK (20 tests, 39 assertions)


$ vendor/bin/paratest --passthru='--repeat 3'
Running phpunit in 4 processes with /var/www/html/vendor/phpunit/phpunit/phpunit

Configuration read from /var/www/html/phpunit.xml

............................................................      60 / 60 (100%)

Time: 00:00.560, Memory: 16.00 MB

OK (60 tests, 117 assertions)
```

stephandesouza:
> And what about simply `--passthru` it like other options like `--group`:
This simplifies a lot! I will make the rollback and adapt the `Option` class

phpstan-baseline.neon
```diff
@@ -94,6 +94,11 @@ parameters:
 			count: 1
 			path: src/Runners/PHPUnit/Options.php
 
+		-
+			message: "#^Parameter \\#37 \\$repeat of class ParaTest\\\\Runners\\\\PHPUnit\\\\Options constructor expects string\\|null, array\\<string\\>\\|bool\\|int\\|string\\|null given\\.$#"
```
Slamdunk, line 98: This shouldn't be needed

src/Runners/PHPUnit/Options.php
```diff
@@ -219,6 +219,8 @@ final class Options
     private $orderBy;
     /** @var int */
     private $randomOrderSeed;
+    /** @var string|null */
```
Slamdunk, line 222:
```suggestion
    /** @var int */
```

src/Runners/PHPUnit/Options.php
```diff
@@ -264,7 +266,8 @@ private function __construct(
         int $verbosity,
         ?string $whitelist,
         string $orderBy,
-        int $randomOrderSeed
+        int $randomOrderSeed,
+        ?string $repeat
```
Slamdunk, line 270:
```suggestion
        int $repeat
```

src/Runners/PHPUnit/Options.php
```diff
@@ -425,6 +430,17 @@ public static function fromConsoleInput(InputInterface $input, string $cwd): sel
...
 
+        if (is_string($options['repeat'])) {
+            if (! is_numeric($options['repeat'])) {
+                throw new InvalidArgumentException(sprintf(
+                    'Option --repeat should have a number value, "%s" given',
+                    $options['repeat']
+                ));
+            }
+
+            $filtered['repeat'] = $options['repeat'];
```
Slamdunk, line 441:
```suggestion
            if ($options['repeat'] !== (string) (int) $options['repeat']) {
                throw new InvalidArgumentException(sprintf(
                    'Option --repeat should have an integer number value, "%s" given',
                    $options['repeat']
                ));
            }

            $filtered['repeat'] = (int) $options['repeat'];
```

src/Runners/PHPUnit/Options.php
```diff
@@ -536,7 +552,8 @@ public static function fromConsoleInput(InputInterface $input, string $cwd): sel
             $verbosity,
             $options['whitelist'],
             $options['order-by'] ?? self::ORDER_DEFAULT,
-            (int) $options['random-order-seed']
+            (int) $options['random-order-seed'],
+            $options['repeat']
```
Slamdunk, line 556:
```suggestion
            (int) $options['repeat']
```

src/Runners/PHPUnit/Options.php
```diff
@@ -739,6 +756,12 @@ public static function setInputDefinition(InputDefinition $inputDefinition): voi
                 InputOption::VALUE_OPTIONAL,
                 'Use a specific random seed <N> for random order'
             ),
+            new InputOption(
+                'repeat',
+                'r',
```
Slamdunk, line 761: I prefer no shortcuts, like PHPUnit:
```suggestion
                null,
```

src/Runners/PHPUnit/Options.php
```diff
@@ -1085,6 +1108,11 @@ public function randomOrderSeed(): int
         return $this->randomOrderSeed;
     }
 
+    public function repeat(): ?string
```
Slamdunk, line 1111:
```suggestion
    public function repeat(): int
```

test/Unit/Runners/PHPUnit/OptionsTest.php
```diff
@@ -296,6 +304,7 @@ public function testProvidedOptions(): void
             '--whitelist' => 'WHITELIST',
             '--order-by' => Options::ORDER_RANDOM,
             '--random-order-seed' => (string) $expected_random_seed,
+            '--repeat' => '2',
```
Slamdunk, line 307:
```suggestion
            '--repeat' => 2,
```

test/Unit/Runners/PHPUnit/OptionsTest.php
```diff
@@ -333,6 +342,7 @@ public function testProvidedOptions(): void
         static::assertSame('WHITELIST', $options->whitelist());
         static::assertSame(Options::ORDER_RANDOM, $options->orderBy());
         static::assertSame($expected_random_seed, $options->randomOrderSeed());
+        static::assertSame('2', $options->repeat());
```
Slamdunk, line 345:
```suggestion
        static::assertSame(2, $options->repeat());
```

Slamdunk (changes requested):

src/Runners/PHPUnit/Options.php
stephandesouza: Will we enforce always sending the repeat option to PHPUnit? If so, we need to set default to '1', no?

src/Runners/PHPUnit/Options.php
stephandesouza: Ooops, okay.

src/Runners/PHPUnit/Options.php
Slamdunk: No: the option in passed only within the `if {}` with the  `$filters` variable.
The `Options::$repeat` attribute is here only as a reference of the CLI arguments

src/Runners/PHPUnit/Options.php
stephandesouza: `$filtered['repeat'] = (int) $options['repeat'];`

will request a chage on $filtered signature:

`* @param array<string, string|int|null> $filtered`

Okay on this?

src/Runners/PHPUnit/Options.php
stephandesouza: Oh, okay.

src/Runners/PHPUnit/Options.php
Slamdunk: Sorry, didn't check that: ok for `$filtered['repeat'] = $options['repeat'];` then.

Slamdunk (approved):

Issue #595, opened 2021-02-18
--teamcity option


neochief:
There's --teamcity option in PHPUnit that changes the default output of PHPUnit from dot format (./F/E/S) to teamcity format. This may seem not that important, but this option is actually used by PHPStorm in its integration with PHPUnit.

I'm 99% sure that having this option working the same way as in PHPUnit would pave the way for PHPStorm integration with paratest (https://youtrack.jetbrains.com/issue/WI-58745).

Slamdunk:
A PR would be very welcome :+1:

neochief:
> A PR would be very welcome 👍

Okay, I'll see what I can do.

MaXal:
@neochief is right! We need this format to integrate ParaTest into PhpStorm. The dot format is not verbose enough to provide us the required information.

See https://github.com/sebastianbergmann/phpunit/blob/master/src/Util/Log/TeamCity.php for the reference.

neochief:
All right, I implemented the basic solution. @Slamdunk , can you please check it out and let me know if it is okay (https://github.com/paratestphp/paratest/pull/597)? I never contributed to Paratest or PHPUnit, so I might be missing something important. I guess I'll need to write some tests as well if the code is okayish.

During manual testing, I tricked PHPStorm by substituting phpunit binary with updated paratest and the IDE actually did it thing with running all tests, so it's a big win.

There are two caveats though:

1. To allow re-executing failed tests from within the IDE, the "filter" option needs to be compatible with PHPUnit. Currently paratest displays the "Option --filter is not implemented for non functional mode" when launching the test the way PHPStorm does this. 

2. To allow running a particular test among the test results, the "--test-suffix" should present and be compatible with PHPUnit. Currently it's not implemented.

I'm guessing these caveats are non-trivial, so I'd like Paratest maintainer's opinion on how to approach the implementation.

arashdalir:
@neochief @Slamdunk 
a question for both of you: wouldn't it be easier to use the `--log-teamcity` feature instead of reimplementing this feature? 

[...]

PR #597, opened 2021-02-24
Initial teamcity output support.

neochief:
See https://github.com/paratestphp/paratest/issues/595

### DIFF
--- a/src/Runners/PHPUnit/Options.php
+++ b/src/Runners/PHPUnit/Options.php
@@ -693,12 +693,6 @@ final class Options
                 'Log test execution in JUnit XML format to file.'
             ),
             new InputOption(
-                'teamcity',
-                null,
-                InputOption::VALUE_NONE,
-                'Output test results in Teamcity format.'
-            ),
-            new InputOption(
                 'log-teamcity',
                 null,
                 InputOption::VALUE_REQUIRED,
@@ -786,6 +780,12 @@ final class Options
                 null,
                 InputOption::VALUE_NONE,
                 'Don\'t start any more processes after a failure.'
+            ),
+            new InputOption(
+                'teamcity',
+                null,
+                InputOption::VALUE_NONE,
+                'Output test results in Teamcity format.'
             ),
             new InputOption(
                 'testsuite',
--- a/src/Runners/PHPUnit/Options.php
+++ b/src/Runners/PHPUnit/Options.php
@@ -209,6 +209,8 @@ final class Options
     private $cwd;
     /** @var string|null */
     private $logJunit;
+    /** @var bool */
+    private $teamcity;
     /** @var string|null */
     private $logTeamcity;
     /** @var string|null */
@@ -249,6 +251,7 @@ final class Options
         bool $functional,
         array $group,
         ?string $logJunit,
+        bool $teamcity,
         ?string $logTeamcity,
         ?int $maxBatchSize,
         bool $noCoverage,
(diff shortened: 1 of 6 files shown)


src/Runners/PHPUnit/Options.php
```diff
@@ -1078,14 +1090,19 @@ public function logJunit(): ?string
...
+
     public function logTeamcity(): ?string
     {
         return $this->logTeamcity;
     }
 
     public function hasLogTeamcity(): bool
     {
-        return $this->logTeamcity !== null;
+        return $this->teamcity || $this->logTeamcity !== null;
```
Slamdunk, line 1105: Output and logging are two different things and shouldn't be coupled

src/Runners/PHPUnit/ResultPrinter.php
```diff
@@ -122,6 +126,9 @@ public function __construct(LogInterpreter $results, OutputInterface $output, Op
         $this->output  = $output;
         $this->options = $options;
 
+        $this->printsTeamcity = $this->options->teamcity();
+        $this->needsTeamcity = $this->printsTeamcity || $this->options->logTeamcity();
```
Slamdunk, line 130: Same here: output and logging are two different things and shouldn't be coupled

Slamdunk (changes requested):
Before reviewing the code, I'd like to review the tests, which are missing, please write the tests before the code

src/Runners/PHPUnit/ResultPrinter.php
neochief: I have moved this check into a method in the Options: https://github.com/paratestphp/paratest/pull/597/files#diff-e7eb5918ab64b6a45aac1cb6586bd195c5a96fd600882a7be2478afe58edbe5cR1108

neochief:
> Before reviewing the code, I'd like to review the tests, which are missing, please write the tests before the code

Okay, I've added a few tests and assertions.

src/Runners/PHPUnit/Options.php
Slamdunk: This point doesn't seem resolved to me: a user could use `--teamcity` but not `--log-teamcity` and viceversa, please split the two behaviors

neochief: Hm, this seems to be already changed, see:
https://github.com/paratestphp/paratest/pull/597/files#diff-e7eb5918ab64b6a45aac1cb6586bd195c5a96fd600882a7be2478afe58edbe5cR1107

The resulting code looks like this:
```
...
public function hasLogTeamcity(): bool
{
    return $this->logTeamcity !== null;
}

public function needsTeamcity(): bool
{
    return $this->teamcity() || $this->hasLogTeamcity();
}
...
```

neochief:
PR can't pass static analysis checks, but it seems that the remaining issues are also present in master.

pyrou:
Any update on this?

Slamdunk:
See https://github.com/paratestphp/paratest/issues/595#issuecomment-799505182

I thank you @neochief for the effort spent in this PR, but this approach doesn't fit the goal; we need a separate binary to serve Paratest to PHPStorm disguised as PHPUnit.

I'm busy right now: any help would be welcome; I'm going to keep this open only as a reminder

Slamdunk:
It was not my intention to close this PR while renamed the default branch.

Nevertheless my last comment is still valid, hope to see someone implementing it

Slamdunk:
Thank you @neochief , this PR has been integrated in https://github.com/paratestphp/paratest/pull/672 and merged

PR #602, opened 2021-03-13
Minor readability updates

shreypuranik:
Some minor updates to aid with readability: 
- Moving private methods lower down into class structure so public methods are shown first


Slamdunk:
Thank you for your commitment, but this kind of changes out of nowhere only throw away the `git-blame` usefullness

Issue #619, opened 2021-08-02
Introduce TESTRUN_TOKEN enviroment variable


remorhaz:
Sometimes we need to guarantee that some setup/teardown action runs exactly once - for example, cleaning the folder before putting there report data.

I'm writing a new integration for Allure reports, and they seem to run great in parallel because each test dumps separate file now - but I experience problems supporting the auto-cleanup option (which should remove all the files from output directory before running first test). My PHPUnit extension hooked `BeforeFirstTest`, but of course Paratest runs it for each process, crushing runner or corrupting results.

Is there any possibility to run some code exactly once before running the tests? For now I'm just switching the auto-cleanup off if TEST_TOKEN is defined, but it would be great to support it for parallel runs.

remorhaz:
In fact, I can use file locks to implement what I need. I have `TEST_TOKEN` and `UNIQUE_TEST_TOKEN` variables, but I see no way to detect if given process runs first time.

That could be solved if Paratest provides a variable with unique value for the whole test run. Something like `TESTRUN_TOKEN` with just `uniqid()` as value; in this case I can do the following in `BeforeFirstTest` hook:

- Try to aquire a lock from a process.
- On failure, do nothing.
- On success, try to read a token from lock file.
- If file not exists or contains no current token, then write current token into file and perform a cleanup.
- Release lock.

What do you think about introducing such a variable?

remorhaz:
I've made a small PR #621 that would solve my problem.

Slamdunk:
Sorry for the late reply.

The issue you are facing can be solved by the native support for ENV variables.
What has been coded in https://github.com/paratestphp/paratest/pull/621 can identically be achieved prepending the env to the binary:
```console
$ TESTRUN_TOKEN=123 vendor/bin/phpunit
$ TESTRUN_TOKEN=123 vendor/bin/paratest
```

If you need a monotonic yet unique value use the current nanoseconds:

```console
$ TESTRUN_TOKEN=$(date +%s.%N) vendor/bin/paratest
```

Since this can be accomplished outside Paratest and we need to take the code complexity low, I'm going to close this.

PR #621, opened 2021-08-03
Test run token added (#619)

remorhaz:


Slamdunk:
See https://github.com/paratestphp/paratest/issues/619#issuecomment-909273971

PR #626, opened 2021-08-17
Respect configuration options of text coverage

### LINKED ISSUE #385: coverage-text in file ?
Hi!

It's not possible to put the coverage in text format inside a file?  
  
Currently, i have the following error:  
`The "--coverage-text" option does not accept a value.`  
  
Thx.

LeoVie:
This resolves #385

`showOnlySummary` was changed to `false` in `test/fixtures/phpunit-fully-configured.xml`, as the other tests rely on the whole text coverage report being created.

I created `test/fixtures/phpunit-coverage-text-show-only-summary.xml`, which only differs from `test/fixtures/phpunit-fully-configured.xml` in part of the option `showOnlySummary` being set to `true`.

For being able to use the dataProvider `generateTextProvider` it was neccessary to overwrite the default configuration in `test/Unit/Coverage/CoverageReporterTest.php`. Therefore I extracted a method `createCoverageReporter` which takes a fixture path as an argument (to reduce code redundancy). This function is called in `setup`, as well as in the test method `testGenerateText`.

With `showOnlySummary` set to false, the text coverage report looks like this:
```
Code Coverage Report:   
  2021-08-17 07:45:37   
                        
 Summary:               
  Classes: 100.00% (1/1)
  Methods: 100.00% (1/1)
  Lines:   100.00% (1/1)

ParaTest\Tests\Unit\Coverage\CoverageReporterTest
  Methods: 100.00% ( 1/ 1)   Lines: 100.00% (  1/  1)
```

With `showOnlySummary` set to true, the text coverage report looks like this:
```
Code Coverage Report Summary:
  Classes: 100.00% (1/1)     
  Methods: 100.00% (1/1)     
  Lines:   100.00% (1/1)
```
So the test checks, if the report contains `Code Coverage Report:`, when `showOnlySummary` is false and `Code Coverage Report Summary:`, when `showOnlySummary` is true.

In lines 103 and 104, the parameters `lowUpperBound` and `highLowerBound` are hard coded to the default parameter values. PHPUnit does not seem to make these parameters configurable for text coverage.

### DIFF
--- a/src/Coverage/CoverageReporter.php
+++ b/src/Coverage/CoverageReporter.php
@@ -99,9 +99,10 @@ final class CoverageReporter
     {
         $text = new Text();
         if ($this->codeCoverageConfiguration !== null && $this->codeCoverageConfiguration->hasText()) {
-            $text = new Text(
-                50,
-                90,
+            $hasHtml = $this->codeCoverageConfiguration->hasHtml();
+            $text    = new Text(
+                $hasHtml ? $this->codeCoverageConfiguration->html()->lowUpperBound() : 50,
+                $hasHtml ? $this->codeCoverageConfiguration->html()->highLowerBound() : 90,
                 $this->codeCoverageConfiguration->text()->showUncoveredFiles(),
                 $this->codeCoverageConfiguration->text()->showOnlySummary()
             );
--- a/src/Coverage/CoverageReporter.php
+++ b/src/Coverage/CoverageReporter.php
@@ -98,6 +98,14 @@ final class CoverageReporter
     public function text(): string
     {
         $text = new Text();
+        if ($this->codeCoverageConfiguration !== null && $this->codeCoverageConfiguration->hasText()) {
+            $text = new Text(
+                50,
+                90,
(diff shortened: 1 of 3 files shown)


LeoVie:
May I ping on this, @Slamdunk @andreasschroth? :) I'm open to improvement suggestions.

src/Coverage/CoverageReporter.php
```diff
@@ -98,6 +98,14 @@ public function php(string $target): void
     public function text(): string
     {
         $text = new Text();
+        if ($this->codeCoverageConfiguration !== null && $this->codeCoverageConfiguration->hasText()) {
+            $text = new Text(
+                50,
+                90,
```
Slamdunk, line 104: PHPUnit gets this values from `coverageReportHtmlType`, can we do the same here and default to 50/90 when those are not set?

LeoVie: If I got you right, it's implemented like this, now.

Slamdunk (approved):

Slamdunk:
Thank you @LeoVie

PR #628, opened 2021-08-25
Fix loading tests

### LINKED ISSUE #627: Paratest running less tests than PHPUnit
| Q | A
| --- | ---
| ParaTest version | 6.3.0
| PHPUnit version | 9.5.8
| PHP version | 7.4.20

#### Summary

When running PHPUnit I get `8063 / 8063 (100%)` in the end - meaning it ran 8063 tests in total.

Now I need to speed it up so I added paratest into my composer.json and replaced `vendor/bin/phpunit` with `vendor/bin/paratest` - fully expecting many tests to fail without further configuration.

What I got however was not expected at all. The tests didn't actually fail, however paratest only got `1654 / 1654 (100%)` meaning most of my tests were completely ignored.

Any idea what could be the cause? I'm running both phpunit and paratest without any parameters. As for phpunit.xml.dist, I have this very simple configuration:

```
  <testsuites>
    <testsuite name="tests">
      <directory suffix=".php">src</directory>
    </testsuite>
  </testsuites>
```

Any idea what could be the cause for Paratest ignoring most of my tests?

enumag:
Closes #627

### DIFF
--- a/src/Runners/PHPUnit/SuiteLoader.php
+++ b/src/Runners/PHPUnit/SuiteLoader.php
@@ -192,16 +192,22 @@ final class SuiteLoader
             // The $class->getParentsCount() + array_merge(...$loadedSuites) stuff
             // are needed to run test with child tests early, because PHPUnit autoloading
             // of such classes in WrapperRunner environments fails (Runner is fine)
-            $loadedSuites = [];
+            $classes = [];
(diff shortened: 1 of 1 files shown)


Slamdunk:
Hi, can you add a test that prove the bug please?

PR #630, opened 2021-08-30
Fix #629 JUnit log, create directory if not existent

### LINKED ISSUE #629: [JUnit/Writer] JUnit report not found in nonexistent directory
| Q | A
| --- | ---
| ParaTest version | 6.3.1
| PHPUnit version | 9.5.8
| PHP version | 8.0.9

#### Summary
It seems the JUnit writer cannot create a report in a nonexistent directory.

#### Current behavior & How to reproduce: (command, code and error stack trace)
for example when running tests like this :

```
XDEBUG_MODE=coverage vendor/bin/paratest -f --log-junit var/coverage/junit.xml 
```

if `var/coverage` doesn't exist, it ends up outputting this:

```
file_put_contents(var/coverage/junit.xml): Failed to open stream: No such file or directory in vendor/brianium/paratest/src/Logging/JUnit/Writer.php on line 99
```

#### Expected behavior

No error. Currently it can be achieved by creating the missing dir. But using phpunit only, there's no such error ...

raneomik:
proposal to #629

### DIFF
--- a/src/Logging/JUnit/Writer.php
+++ b/src/Logging/JUnit/Writer.php
@@ -10,20 +10,18 @@ use ParaTest\Logging\LogInterpreter;
 
 use function assert;
 use function count;
-use function file_exists;
+use function dirname;
 use function file_put_contents;
 use function get_object_vars;
 use function htmlspecialchars;
+use function is_dir;
 use function is_float;
 use function is_scalar;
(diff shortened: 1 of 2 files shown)


Slamdunk:
Thank you @raneomik

PR #659, opened 2022-03-28
Fix detection of logical CPU cores on windows

staabm:
before this change only physical cpus were detected on windows.
initially reported in https://github.com/phpstan/phpstan-src/pull/1115#issuecomment-1079862741

fix inspired by similar code in phpstan-src:
https://github.com/phpstan/phpstan-src/blob/a95ce30d8d036fe5232815ff12d229cea4b1f46a/src/Process/CpuCoreCounter.php#L41-L50

I ran a few benchmarks, and interesstingly the overall duration when running the phpstan test-suite before/after this PR did nearly not change. 
nevertheless I figured this fixed should be merged, as it adjusts the implementation to the actual expectations described in the method body.

### DIFF
--- a/src/Runners/PHPUnit/Options.php
+++ b/src/Runners/PHPUnit/Options.php
@@ -883,7 +883,7 @@ final class Options
         // @codeCoverageIgnoreStart
         } elseif (DIRECTORY_SEPARATOR === '\\') {
             // Windows
-            if (($process = @popen('wmic cpu get NumberOfCores', 'rb')) !== false) {
+            if (($process = @popen('wmic cpu get NumberOfLogicalProcessors', 'rb')) !== false) {
                 fgets($process);
                 $cores = (int) fgets($process);
                 pclose($process);


ondrejmirtes:
Oh, could you send a patch to phpstan-src to this directory? https://github.com/phpstan/phpstan-src/tree/1.5.x/patches should make Windows build faster again :)

Slamdunk:
> I ran a few benchmarks, and interestingly the overall duration when running the phpstan test-suite before/after this PR did nearly not change.

I acknowledged a year ago the `NumberOfCores-to-NumberOfLogicalProcessors` tweak PHPStan did to the code copied from this repo, but never bothered applying it here because I've never seen an perceptible improvement.

Btw I'm not an expert on Windows CPU management, so I trust you this is the way to go.

Slamdunk (approved):

Slamdunk:
> Oh, could you send a patch to phpstan-src to this directory? https://github.com/phpstan/phpstan-src/tree/1.5.x/patches should make Windows build faster again :)

Already released in [`v6.4.4`](https://github.com/paratestphp/paratest/releases/tag/v6.4.4) so I think no need for a patch

ondrejmirtes:
Awesome, you're fast! I just updated it in phpstan-src, so we'll see :)

PR #666, opened 2022-06-15
CS: disable mysterious Squiz.Functions.FunctionDeclaration.Found issue

Slamdunk:


composer.json
```diff
@@ -38,22 +38,22 @@
         "ext-pcre": "*",
         "ext-reflection": "*",
         "ext-simplexml": "*",
-        "phpunit/php-code-coverage": "^9.2.11",
+        "phpunit/php-code-coverage": "^9.2.15",
```
mfn, line 41: Stupid question, but I guess it is **required** to have 9.2.15? 

In a private project I'm using paratest + infection, but there's a coverage bug with 9.2.15 so I had to pin it to an earlier version and now I have this conflict 😅

Slamdunk: > Stupid question

There's no stupid question, only stupid answers :wink: 

> I guess it is **required** to have 9.2.15

No it isn't.
In [this particular project](https://github.com/paratestphp/paratest#versions) I needed many times to pinpoint exact patch releases for dependencies, and then I get used to do it every time to avoid nightmares.
I this case there wasn't a need, but I did it anyway.

> In a private project I'm using paratest + infection, but there's a coverage bug with 9.2.15 so I had to pin it to an earlier version and now I have this conflict sweat_smile

I understand your case and I'm sorry for that, but I won't change it.
Please stick to a previous version ([`v6.4.4`](https://github.com/paratestphp/paratest/releases/tag/v6.4.4)) and cherry-pick what you need with [`cweagans/composer-patches`](https://github.com/cweagans/composer-patches)

mfn: Can't beat honest answers, thank you for the insights :-}

PR #668, opened 2022-06-15
Add version information

### LINKED ISSUE #624: `--version` does not return SemVer version of neither Paratest nor PHPUnit
| Q | A
| --- | ---
| ParaTest version | 6.3.1
| PHPUnit version | 9.5.8
| PHP version | 8.0.8

Probably related to https://github.com/paratestphp/paratest/pull/481

#### Summary

Running `paratest --version` returns `Console Tool`. It breaks any version parsers and does not provide any helpful information about what Paratest / underlying PHPUnit is used.

To understand the version, I need to do `composer show | grep paratest`.

In additional to that it's not convenient, we also need this in @infection. I'm working on Paratest adapter in order to make it work with Infection.

Under the hood, Infection relies on PHPUnit's version to understand what feature can be used in the generated `phpunit.xml` files. 

#### Current behavior

`paratest --version` returns `Console Tool`

#### Expected behavior

`paratest --version` returns, for example `PHPUnit 9.5.8, Paratest 6.5.3`. If it's not possible, returning `Paratest 6.5.3` would still be helpful.

Jean85:
Fixes #624.

This leverages my own package that isolate us from any transient dependency: https://github.com/jean85/pretty-package-versions#compatibility

The output with this PR is:
immagine

With a tagged version, the output would be:
```
$ vendor/bin/paratest --version                                                                 
Paratest 6.4.0
```

### DIFF
--- a/composer.json
+++ b/composer.json
@@ -38,6 +38,7 @@
         "ext-pcre": "*",
         "ext-reflection": "*",
         "ext-simplexml": "*",
+        "jean85/pretty-package-versions": "^1.6 || ^2.0",
         "phpunit/php-code-coverage": "^9.2.15",
         "phpunit/php-file-iterator": "^3.0.6",
         "phpunit/php-timer": "^5.0.3",
--- a/composer.json
+++ b/composer.json
@@ -38,7 +38,7 @@
         "ext-pcre": "*",
         "ext-reflection": "*",
         "ext-simplexml": "*",
-        "jean85/pretty-package-versions": "^1.6 || ^2.0",
+        "jean85/pretty-package-versions": "^2.0",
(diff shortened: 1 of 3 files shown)


composer.json
```diff
@@ -38,6 +38,7 @@
         "ext-pcre": "*",
         "ext-reflection": "*",
         "ext-simplexml": "*",
+        "jean85/pretty-package-versions": "^1.6 || ^2.0",
```
Slamdunk: IIRC, `v1.6` is needed for `composer:v1` only.
Given the [steady decline](https://packagist.org/packages/jean85/pretty-package-versions/stats) of `v1` in favor of `v2`, I'd say let's push `v2` only :rocket:

Slamdunk (approved):

PR #686, opened 2022-08-10
Allows phpstormhelper to handle and account for coverage options

### LINKED ISSUE #685: Coverage with phpstorm
| Q | A
| --- | ---
| ParaTest version | 6.6.1
| PHPUnit version | 9.5.21
| PHP version | 8.1.8

#### Summary

When running with coverage, it throw an error, that the process argument is not found

#### Current behavior

When running it runs with phpunit and not paratest

Create a phpstorm convide as described in the readme, and then run with coverage

#### Expected behavior

Should run with coverage

The reason why it's doing it is because when preparing, paratest_for_phpstorm removes the phpunit command line argument. But when running with coverage, phpstorm adds an aditional argument to enable pcov.

I managed to make an hotfix for this but I don't think it's clean enough but it may be a start, in src/Util/PhpstormHelper.php, I added at line 21 :
```php
if($argv[1] === '-dpcov.enabled=1') {
    unset($argv[1]);
    unset($argv[2]);

    return $paratestBinary;
}
```

billypoke:
Closes #685

### DIFF
--- a/test/Unit/Util/PhpstormHelperTest.php
+++ b/test/Unit/Util/PhpstormHelperTest.php
@@ -9,6 +9,7 @@ use ParaTest\Util\PhpstormHelper;
 use PHPUnit\Framework\TestCase;
 
 use function array_values;
+use function sprintf;
 use function uniqid;
 
 /**
@@ -33,14 +34,14 @@ final class PhpstormHelperTest extends TestCase
         $actualBinary = PhpstormHelper::handleArgvFromPhpstorm($argv, $paratestBinary);
         $argv         = array_values($argv);
 
-        static::assertSame($expectedArgv, $argv);
+        self::assertSame($expectedArgv, $argv);
         self::assertSame($expectedBinary, $actualBinary);
     }
 
     public function providePhpstormCases(): Generator
     {
-        $paratestBinary = uniqid('paratest_');
-        $phpunitBinary  = uniqid('phpunit_');
+        $paratestBinary = sprintf('%s/vendor/brianium/paratest/bin/paratest', uniqid());
+        $phpunitBinary  = sprintf('%s/vendor/phpunit/phpunit/phpunit', uniqid());
 
         $argv   = [];
         $argv[] = $paratestBinary;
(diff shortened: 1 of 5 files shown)


test/Unit/Util/PhpstormHelperTest.php
```diff
@@ -94,5 +94,76 @@ public function providePhpstormCases(): Generator
...
+            $paratestBinary,
+            $paratestBinary,
+        ];
+
+        $argv   = [];
+        $argv[] = $paratestBinary;
+        $argv[] = $phpunitBinary;
+        $argv[] = '--runner';
+        $argv[] = 'WrapperRunner';
+        $argv[] = '-dpcov.enabled=1';
```
Slamdunk: This isn't the order I get in my PHPStorm:

```php
array (
  0 => './bin/paratest_for_phpstorm',
  1 => '-dpcov.enabled=1',
  2 => '/home/tessarotto/repos/paratest/vendor/phpunit/phpunit/phpunit',
  3 => '--runner',
  4 => 'WrapperRunner',
  5 => '--coverage-clover',
  6 => '/home/tessarotto/.cache/JetBrains/PhpStorm2022.2/coverage/paratest@ParaTest.xml',
  7 => '--configuration',
  8 => '/home/tessarotto/repos/paratest/phpunit.xml',
  9 => '--teamcity',
)
```

Can you add this to data sets?
At this point the original `unset($argv[1]);` is not enough, we need to cycle all the `$argv` var and remove the first entry that ends with `/phpunit`

Slamdunk (changes requested):
Hi, thank you for this proposal.

I feel that this doesn't solve the underlying issue though. PhpStorm adds `-dpcov.enabled=1` to be sure PCOV is enabled, that indeed might be disabled in the system configuration.
Removing the flag here is necessary, but PCOV or xDebug are still disabled in such cases.
We should edit the `README.md` adding that:

1. It is suggested to already have PCOV/xDebug enabled for code coverage
2. When it's not a viable, add a second Configuration with `-dpcov.enabled=1`/`'-dxdebug.mode=coverage'` prepended to the `./vendor/bin/paratest_for_phpstorm` interpreter option

test/Unit/Util/PhpstormHelperTest.php
billypoke: Ok, the test cases are updated to match, the order of arguments on my machine matches yours. I also switched to searching for the necessary argv keys to unset instead of hard-coding the array keys. Let me know if this is what you were aiming for :)

Slamdunk (approved):

PR #714, opened 2022-12-22
WrapperRunner: add `--max-batch-size` support

### LINKED ISSUE #713: Support batchsize for WrapperRunner
Hello,

With my team, we had intermittent problems with memory leaks on a symfony test suite. We found that some memory was stuck on each worker, and depending on the tests distribution, sometimes the paratest crashed.

We found that resetting the workers sometime stopped the problem. Is that something you would be willing to support ?

If so, I have prepared a pull request

CheapHasz:
Fix  https://github.com/paratestphp/paratest/issues/713

I have found that the WrapperWorker did not issue a stop command to the process, which left it hanging. I suppose we could wait for the exit command to be resolved, but it takes some avoidable time.

I also might have gone overboard with the worker destruction, maybe a maintainer will have a more subtle approach to suggest.

I am more than willing to fix the mutation tests, but i am not quite sure how to do it.

### DIFF
--- a/src/Runners/PHPUnit/Worker/WrapperWorker.php
+++ b/src/Runners/PHPUnit/Worker/WrapperWorker.php
@@ -166,7 +166,6 @@ final class WrapperWorker
 
     public function stop(): void
     {
-        $this->process->stop();
         $this->input->write(self::COMMAND_EXIT);
     }
 
--- a/src/Runners/PHPUnit/Worker/WrapperWorker.php
+++ b/src/Runners/PHPUnit/Worker/WrapperWorker.php
@@ -166,6 +166,7 @@ final class WrapperWorker
 
     public function stop(): void
     {
+        $this->process->stop();
         $this->input->write(self::COMMAND_EXIT);
     }
 
--- a/src/Runners/PHPUnit/WrapperRunner.php
+++ b/src/Runners/PHPUnit/WrapperRunner.php
@@ -17,8 +17,11 @@ use function usleep;
 /** @internal */
 final class WrapperRunner extends BaseRunner
 {
-    /** @var WrapperWorker[] */
+    /** @var array<int,WrapperWorker> */
     private $workers = [];
+
+    /** @var array<int,int> */
+    private $batches = [];
 
     protected function beforeLoadChecks(): void
     {
@@ -40,8 +43,7 @@ final class WrapperRunner extends BaseRunner
     private function startWorkers(): void
     {
         for ($token = 1; $token <= $this->options->processes(); ++$token) {
-            $this->workers[$token] = new WrapperWorker($this->output, $this->options, $token);
-            $this->workers[$token]->start();
+            $this->startWorker($token);
         }
     }
 
@@ -49,9 +51,10 @@ final class WrapperRunner extends BaseRunner
     {
         $phpunit        = $this->options->phpunit();
         $phpunitOptions = $this->options->filtered();
+        $batchSize      = $this->options->maxBatchSize();
 
         while (count($this->pending) > 0 && count($this->workers) > 0) {
-            foreach ($this->workers as $worker) {
+            foreach ($this->workers as $token => $worker) {
                 if (! $worker->isRunning()) {
                     throw $worker->getWorkerCrashedException();
                 }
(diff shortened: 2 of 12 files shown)


src/Runners/PHPUnit/Worker/WrapperWorker.php
```diff
@@ -166,6 +166,7 @@ public function reset(): void
 
     public function stop(): void
     {
+        $this->process->stop();
```
Slamdunk: Please propose this tweak in another PR, I have to think about it

src/Runners/PHPUnit/WrapperRunner.php
```diff
@@ -17,9 +17,12 @@
 /** @internal */
 final class WrapperRunner extends BaseRunner
 {
-    /** @var WrapperWorker[] */
+    /** @var array<int,WrapperWorker> */
```
Slamdunk, line 20:
```suggestion
    /** @var array<positive-int,WrapperWorker> */
```

src/Runners/PHPUnit/WrapperRunner.php
```diff
@@ -17,9 +17,12 @@
...
     private $workers = [];
 
+    /** @var array<int,int> */
+    private $batches = [];
```
Slamdunk, line 24:
```suggestion
    /** @var array<positive-int,int> */
    private array $batches = [];
```

src/Runners/PHPUnit/WrapperRunner.php
```diff
@@ -61,10 +64,16 @@ private function assignAllPendingTests(): void
                 }
 
                 $this->flushWorker($worker);
+                if ($batchSize && $this->batches[$token] > $batchSize) {
```
Slamdunk:
```suggestion
                if (null !== $batchSize && $this->batches[$token] === $batchSize) {
```
If you use the `>` operator you are actually executing `maxBatchSize + 1` tests in each batch.
`=== ` is needed

src/Runners/PHPUnit/WrapperRunner.php
```diff
@@ -125,4 +134,18 @@ private function waitForAllToFinish(): void
...
+        $this->workers[$token] = new WrapperWorker($this->output, $this->options, $token);
+        $this->workers[$token]->start();
+        $this->batches[$token] = 0;
+    }
+
+    private function destroyWorker(int $token): void
+    {
+        $this->workers[$token]->stop();
+        $this->workers[$token]->__destruct();
+        unset($this->workers[$token]);
```
Slamdunk:
```suggestion
        unset($this->workers[$token]);
```
`$this->workers` is the only reference to the object, so the `unset()` should be enough

test/Unit/Runners/PHPUnit/WrapperRunnerTest.php
```diff
@@ -42,4 +42,13 @@ public function testWrapperRunnerWorksWellWithManyTests(): void
...
         $this->runRunner();
     }
+
+    public function testWrapperRunnerHandlesBatchSize(): void
+    {
+        $this->bareOptions['--path']           = $this->fixture('passing_tests' . DS . 'level1' . DS . 'level2');
+        $this->bareOptions['--configuration']  = $this->fixture('phpunit-parallel-suite-with-globals.xml');
+        $this->bareOptions['--max-batch-size'] = 2;
+
+        $this->runRunner();
```
Slamdunk, line 74: This is not a test, it's just a mere execution of something.

A proper test for the `maxBatchSite` should:

1. Consist of 4 separate tests
2. Run in 2 processes
3. Each test should track both `getmypid()` and `TEST_TOKEN` env var unique values
4. Assert that `getmypid()` unique values are `4`, and `TEST_TOKEN` unique values are `2`

Slamdunk (changes requested):
Thank you for the PR. I've added some notes.

Please also update the docs: https://github.com/paratestphp/paratest#optimizing-speed

src/Runners/PHPUnit/Worker/WrapperWorker.php
CheapHasz: I was under the impression that processes were dangling without that, but you're right, they eventually are removed

test/Unit/Runners/PHPUnit/WrapperRunnerTest.php
CheapHasz: I didn't think about using the testcases that way.

I think I did come up with something useful

PR #724, opened 2023-02-04
Fix fidry/cpu-core-counter version

samsonasik:
The `||` used in equal value `^0.5.1 || ^0.5.1`, so it can just use the value directly.

### DIFF
--- a/composer.json
+++ b/composer.json
@@ -38,7 +38,7 @@
         "ext-pcre": "*",
         "ext-reflection": "*",
         "ext-simplexml": "*",
-        "fidry/cpu-core-counter": "^0.5.1",
+        "fidry/cpu-core-counter": "^0.4.1 || ^0.5.1",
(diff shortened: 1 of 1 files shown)


Slamdunk:
Uh, `composer bump` upgraded the first pin of 0.4 to 0.5 instead of leaving there both 🤦

samsonasik:
Should I update to ^0.4.0 || ^0.5.1?

Slamdunk:
Yes please: ^0.4.1 || ^0.5.1

samsonasik:
Done 👍

Slamdunk:
Thank you @samsonasik

Jean85:
> Uh, `composer bump` upgraded the first pin of 0.4 to 0.5 instead of leaving there both facepalm

`bump` shouldn't be used on distributed libraries, only on end projects.

Issue #742, opened 2023-03-13
Output broken in 7.1.1


dan-the-dev:
| Q | A
| --- | ---
| ParaTest version | 7.1.1
| PHPUnit version | 10.0.16
| PHP version | 8.2.0

#### Summary

After the fix for PHPUnit 10.0.16 compatibility, tests are back to green but the output is weird, it reaches 775% like if tests are executed multiple times. 

#### Current behavior

Output is weird:

```
ParaTest v7.1.1 upon PHPUnit 10.0.16 by Sebastian Bergmann and contributors.

                                Processes:     4
Runtime:       PHP 8.2.0
Configuration: /home/runner/work/my-bistrot-backend/my-bistrot-backend/phpunit.xml

        .        .        .        .        .        .          61 / 1647 (  3%)
 .        .        .        .        .        .        .       122 / 1647 (  7%)
   .        .        .        .        .        .        .     183 / 1647 ( 11%)
     .        .        .        .        .        .        .   244 / 1647 ( 14%)
       .        .        .        .        .        .          305 / 1647 ( 18%)
.        .        .        .        .        .        .        366 / 1647 ( 22%)
  .        .        .        .        .        .        .      427 / 1647 ( 25%)
    .        .        .        .        .        .        .    488 / 1647 ( 2[9](https://github.com/mymenudev/my-bistrot-backend/actions/runs/4403898009/jobs/7712831634#step:8:10)%)
      .        .        .        .        .        .        .  549 / 1647 ( 33%)
        .        .        .        .        .        .         6[10](https://github.com/mymenudev/my-bistrot-backend/actions/runs/4403898009/jobs/7712831634#step:8:11) / 1647 ( 37%)
 .        .        .        .        .        .        .       671 / 1647 ( 40%)
   .        .        .        .        .        .        .     732 / 1647 ( 44%)
     .        .        .        .        .        .        .   793 / 1647 ( 48%)
       .        .        .        .        .        .          854 / 1647 ( 51%)
.        .        .        .        .        .        .        915 / 1647 ( 55%)
  .        .        .        .        .        .        .      976 / 1647 ( 59%)
    .        .        .        .        .        .        .   1037 / 1647 ( 62%)
      .        .        .        .        .        .        . 1098 / 1647 ( 66%)
        .        .        .        .        .        .        [11](https://github.com/mymenudev/my-bistrot-backend/actions/runs/4403898009/jobs/7712831634#step:8:12)59 / 1647 ( 70%)
[...]

PR #755, opened 2023-05-02
Implement dataset parallelization with `-f` flag

danog:
Brings back the feature that was removed in the phpunit 10 update.

### DIFF
--- a/composer.json
+++ b/composer.json
@@ -58,7 +58,8 @@
         "phpstan/phpstan-phpunit": "^1.3.11",
         "phpstan/phpstan-strict-rules": "^1.5.1",
         "squizlabs/php_codesniffer": "^3.7.2",
-        "symfony/filesystem": "^6.2.10"
+        "symfony/filesystem": "^6.2.10",
+        "vimeo/psalm": "^5.11"
     },
     "autoload": {
         "psr-4": {
--- a/src/WrapperRunner/SuiteLoader.php
+++ b/src/WrapperRunner/SuiteLoader.php
@@ -28,7 +28,6 @@ use function is_string;
 use function mt_srand;
 use function ob_get_clean;
 use function ob_start;
-use function str_replace;
 use function str_starts_with;
 use function strlen;
 use function substr;
@@ -82,7 +81,7 @@ final class SuiteLoader
             $files[$file] = null;
 
             if ($test instanceof TestCase) {
-                $name = str_replace(' with data set ', '', $test->nameWithDataSet());
+                $name = $test->nameWithDataSet();
             } else {
                 $name = $test->getName();
             }
--- a/src/WrapperRunner/SuiteLoader.php
+++ b/src/WrapperRunner/SuiteLoader.php
@@ -4,6 +4,7 @@ declare(strict_types=1);
 
 namespace ParaTest\WrapperRunner;
 
+use Generator;
 use ParaTest\Options;
 use PHPUnit\Framework\TestCase;
 use PHPUnit\Framework\TestSuite;
@@ -27,6 +28,7 @@ use function is_string;
 use function mt_srand;
 use function ob_get_clean;
 use function ob_start;
+use function str_replace;
 use function str_starts_with;
 use function strlen;
 use function substr;
@@ -35,8 +37,11 @@ use function substr;
 final class SuiteLoader
 {
     public readonly int $testCount;
+    public readonly int $testCountWithDatasets;
     /** @var list<non-empty-string> */
     public readonly array $files;
+    /** @var list<non-empty-string> */
+    public readonly array $tests;
 
     public function __construct(
         private readonly Options $options,
@@ -73,8 +78,21 @@ final class SuiteLoader
         $this->testCount = count($testSuite);
 
         $files = [];
-        $this->loadFiles($testSuite, $files);
+        $tests = [];
+        foreach ($this->loadFiles($testSuite) as $file => $test) {
+            $files[$file] = true;
+
+            if ($test instanceof TestCase) {
+                $name = str_replace(' with data set ', '', $test->nameWithDataSet());
+            } elseif ($test instanceof PhptTestCase) {
+                $name = $test->getName();
+            }
+
+            $tests[] = "$file\0$name";
+        }
+
         $this->files = array_keys($files);
+        $this->tests = $tests;
 
         if (! $this->options->configuration->hasCoverageReport()) {
             return;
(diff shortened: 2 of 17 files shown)


src/WrapperRunner/SuiteLoader.php
```diff
@@ -73,8 +77,21 @@ public function __construct(
         $this->testCount = count($testSuite);
 
         $files = [];
-        $this->loadFiles($testSuite, $files);
+        $tests = [];
+        foreach ($this->loadFiles($testSuite) as $file => $test) {
+            $files[$file] = null;
+
+            if ($test instanceof TestCase) {
+                $name = str_replace(' with data set ', '', $test->nameWithDataSet());
```
Slamdunk: Unfortunately this doesn't work for both numeric and string data names.
`PHPUnit\Runner\Filter\Factory` needs:

1. `testName#1` for numeric names
2. `testName with data set "foo"` for string names

test/TestBase.php
```diff
@@ -64,33 +65,63 @@ final protected function createOptionsFromArgv(array $argv, ?string $cwd = null)
...
-            unset($_SERVER[Options::ENV_KEY_UNIQUE_TOKEN]);
-        }
-
-        $exitCode = $runner->run();
-        if ($shouldPutEnvForParatestTestingItSelf) {
-            putenv(Options::ENV_KEY_TOKEN . '=' . $prevToken);
-            putenv(Options::ENV_KEY_UNIQUE_TOKEN . '=' . $prevUniqueToken);
-            $_SERVER[Options::ENV_KEY_TOKEN]        = $prevToken;
-            $_SERVER[Options::ENV_KEY_UNIQUE_TOKEN] = $prevUniqueToken;
+        $last = null;
```
Slamdunk: Please restore the original code here, and add a new dedicated test for the `--functional` option, with both numeric data names and string data names

Slamdunk (changes requested):
Thank you for proposing this. A couple of changes are required :muscle:

Guite:
Looking forward to this one.

cuberinooo:
Would love to see this feature back again :)

Slamdunk:
Help @danog resolving the requested changes and we are ready to go :muscle:

src/WrapperRunner/SuiteLoader.php
Guite: looks like we could just use `$name = $test->nameWithDataSet());`

danog:
Whoops sorry completely forgot about this PR, will apply the required changes this weekend!

test/fixtures/function_parallelization_tests/FunctionalParallelizationTest.php
```diff
@@ -0,0 +1,43 @@
...
+
+namespace ParaTest\Tests\fixtures\paratest_only_tests;
+
+use PHPUnit\Framework\TestCase;
+
+/** @internal */
+final class FunctionalParallelizationTest extends TestCase
+{
+    /** @return list<array{string, string}> */
+    public function dataProvider1(): array
```
Slamdunk: Fixing these two data providers to `static` as required by PHPUnit 10 should be enough to get the tests pass.

composer.json
```diff
@@ -58,7 +58,8 @@
         "phpstan/phpstan-phpunit": "^1.3.11",
         "phpstan/phpstan-strict-rules": "^1.5.1",
         "squizlabs/php_codesniffer": "^3.7.2",
-        "symfony/filesystem": "^6.2.10"
+        "symfony/filesystem": "^6.2.10",
+        "vimeo/psalm": "^5.11"
```
Slamdunk, line 62: I've removed `psalm` fro ParaTest 7 because of a few incompatibilites with `PHPStan` with zero gains: is there a valid reason to re-introduce it? If not I'd prefer to not have it

Slamdunk (changes requested):

composer.json
danog: Psalm is quite a lot faster than phpstan, and it offers many additional features like sealed shapes, list types, and even security analysis (though not really applicable in this project).

I've mainly added it because it's quite a lot faster than phpstan, and during iteration I found it easier to just run psalm instead of phpstan :)

Guite:
If you think this is feature complete please leave a comment. I will then re-test it and report our experience.

danog:
Yeah it's ready, I just can't figure out why are the tests failing (with different errors when running on local, too)

src/WrapperRunner/SuiteLoader.php
Slamdunk: @Guite no: `nameWithDataSet` outputs `testDataProvider1 with data set #0` for numeric names, and that doesn't work with `--filter`

test/Unit/WrapperRunner/WrapperRunnerTest.php
```diff
@@ -523,4 +523,13 @@ public function testProcessIsolation(): void
...
         self::assertEquals(0, $runnerResult->exitCode);
     }
+
+    public function testFunctionalParallelization(): void
+    {
+        $this->bareOptions['path'] = $this->fixture('function_parallelization_tests' . DIRECTORY_SEPARATOR . 'FunctionalParallelizationTest.php');
+        $this->bareOptions['-f']   = true;
+
+        $runnerResult = $this->runRunner();
+        self::assertEquals(0, $runnerResult->exitCode);
```
Slamdunk, line 533: Currently this test is poorly written: executing `./bin/paratest --no-configuration -f test/fixtures/function_parallelization_tests/FunctionalParallelizationTest.php` only runs 3 tests instead of 6, because of https://github.com/paratestphp/paratest/pull/755#discussion_r1206423388

Please:

1. Add a specific assertion on the number of tests expected to be run, six
2. Compose the test name with `\PHPUnit\Framework\TestCase::dataName` API, and use `#` for `int` and `@` for `string`

Slamdunk (changes requested):

oprudkyi:
Hi @danog 

I've tried to check PR, it seems the count is wrong 
```
..........................................                        42 / 42 (100%)
. 43 / 42 (102%)
. 44 / 42 (104%)
. 45 / 42 (107%)
. 46 / 42 (109%)
```

Slamdunk:
@oprudkyi maybe see https://github.com/paratestphp/paratest/issues/742

Slamdunk:
This PR needed some few more tweaks that I addressed in #770 (I could have done it here, if the changes were presented as a separate branch in @danog  fork):

1. Fix for https://github.com/paratestphp/paratest/pull/755#discussion_r1206426509
2. A fix for `PHPT` files
3. A workaround for https://github.com/paratestphp/paratest/pull/770#issuecomment-1590945823

I've also removed the support for `-f` shortcut, leaving `--functional` only because `-f` may be used in the future by PHPUnit.

But all in all, thank you for this PR :muscle:

PR #765, opened 2023-05-25
Add `--stop-on-error` option to runner

### LINKED ISSUE #764: Stop on error?
| Q | A
| --- | ---
| ParaTest version | 6.9.1
| PHPUnit version | 9.6.7
| PHP version | 8.2

#### Summary

Is there a reason why phpunit's `--stop-on-error` is not implemented? Would speed up debugging on test sets that contains an error... right now I have to wait until all tests are done to see the error.

temp:
Resolves #764

### DIFF
--- a/src/Runners/PHPUnit/Runner.php
+++ b/src/Runners/PHPUnit/Runner.php
@@ -91,7 +91,7 @@ final class Runner extends BaseRunner
         }
 
         $this->exitcode = max($this->exitcode, (int) $worker->stop());
-        if ($this->options->stopOnFailure() && $this->exitcode > 0) {
+        if (($this->options->stopOnFailure() || $this->options->stopOnError()) && $this->exitcode > 0) {
             $this->pending = [];
         }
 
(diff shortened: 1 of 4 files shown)


src/Runners/PHPUnit/Runner.php
```diff
@@ -91,7 +91,7 @@ private function testIsStillRunning(RunnerWorker $worker): bool
         }
 
         $this->exitcode = max($this->exitcode, (int) $worker->stop());
-        if ($this->options->stopOnFailure() && $this->exitcode > 0) {
+        if (($this->options->stopOnFailure() || $this->options->stopOnError()) && $this->exitcode > 0) {
```
Slamdunk, line 94: How can `--stop-on-error` be more useful than `--stop-on-failure` if it does the same thing?

Slamdunk (changes requested):

src/Runners/PHPUnit/Runner.php
temp: We have a case where we only want to stop on errors, not failures AND errors. In phpunit 9.x phpunit stop for failures and errors with `--stop-on-failure`, and only on errors with `--stop-on-errors`.
Both options are passed to phpunit, and are handled differently there, paratest can handle both in the same way.

Slamdunk (approved):

PR #770, opened 2023-06-14
Implement dataset parallelization with `--functional` flag

Slamdunk:
Fix https://github.com/paratestphp/paratest/pull/755


Slamdunk:
https://github.com/sebastianbergmann/phpunit/blob/3c86e7ede2b0682097ea4f628b631c30aaee4b00/src/Runner/Filter/NameFilterIterator.php#L60

NULL-byte not accepted by `preg_match` in PHP < 8.2:

1. https://3v4l.org/iENgn
2. https://bugs.php.net/bug.php?id=77726
3. https://github.com/php/php-src/pull/8114

PR #781, opened 2023-07-25
Allow Symfony 7

jmsche:
Hi,

This PR allows packages relying on Paratest to upgrade their dependencies to Symfony 7.

I added a separate workflow section to test specifically against Symfony 7 on Paratest, and it seems to pass.


Slamdunk:
Thank you for your contribution, but I cannot put my energies into software that's not yet released.

Don't worry: we have [dependabot](https://github.com/paratestphp/paratest/blob/7.x/.github/dependabot.yml) active here, as soon as Symfony 7 will be released a new PR will notify me that I have to support it from then on.

PR #794, opened 2023-10-04
Set `PHPUNIT_COMPOSER_INSTALL` before requiring the autoloader so it can access it

### LINKED ISSUE #793: `SYMFONY_DEPRECATIONS_HELPER` has no effect when using the `WrapperRunner`
| Q | A
| --- | ---
| ParaTest version | 6.10.0
| PHPUnit version | 9.6.12
| PHP version | 8.2.8

#### Summary

I wanted to try `SYMFONY_DEPRECATIONS_HELPER`’s `logFile` option, but it only worked when I removed the `--runner=WrapperRunner` option from `paratest`’s CLI.

In your PhpUnit configuration file, add or update `SYMFONY_DEPRECATIONS_HELPER` as such:

```xml
<server name="SYMFONY_DEPRECATIONS_HELPER" value="logFile=var/log/deprecations.test.log" />
```

then run eg. `vendor/bin/paratest --runner=WrapperRunner` (I confirmed the `--processes` option does not make any difference).

#### Current behavior

The `var/log/deprecations.test.log` file is not created.

#### Expected behavior

The `var/log/deprecations.test.log` file is created and filled with collected deprecations.

MatTheCat:
Fixes https://github.com/paratestphp/paratest/issues/793

Some guidance would be appreciated if we want to test this fix..!

### DIFF
--- a/bin/phpunit-wrapper.php
+++ b/bin/phpunit-wrapper.php
@@ -15,8 +15,8 @@ use ParaTest\Runners\PHPUnit\Worker\WrapperWorker;
 
     foreach ($composerAutoloadFiles as $file) {
         if (file_exists($file)) {
-            require_once $file;
             define('PHPUNIT_COMPOSER_INSTALL', $file);
+            require_once $file;
 
             break;
         }


MatTheCat:
@Slamdunk will you include this in a v6.10.1?

Slamdunk:
I already did https://github.com/paratestphp/paratest/releases/tag/v6.10.1

MatTheCat:
Nice, didn’t saw it in the releases page :ok_hand:

PR #844, opened 2024-03-12
Fix problem with untrimmed command

bambamboole:
After update to PHP 8.3 I got the following warnings in the `bin/phpunit-wrapper.php`:
```
Warning: unserialize(): Extra data starting at offset 534 of 535 bytes in /Users/xxx/Projects/xxx/vendor/brianium/paratest/bin/phpunit-wrapper.php on line 41
```

Trimming the command helped there.

### DIFF
--- a/.github/workflows/integrate.yaml
+++ b/.github/workflows/integrate.yaml
@@ -106,6 +106,7 @@ jobs:
           - "8.0"
           - "8.1"
           - "8.2"
+          - "8.3"
         dependencies:
           - "lowest"
           - "highest"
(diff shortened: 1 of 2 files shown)


Slamdunk:
Can you add a test please?

bambamboole:
@Slamdunk , maybe you can support me here, because I am having a hard time reproducing this in a test case. First of all I have extended now the CI test suite to also run on PHP 8.3. But how can I assert such a case in a bast way? 
I "believe", that the error in general comes from the case, that PHP8.3 throws a warning like `Warning: unserialize(): Extra data star...` and because of that additional output the execution fails.

Slamdunk:
I see only now that this PR targets `6.x` branch: it's EOLed and I don't care much anymore, so I'll merge and release this as is.

Thank you for your contribution.

PR #867, opened 2024-08-05
Fix `LogMerger` when no log is created

CanvasCompanyHylke:
When thare are no tests to run, the `LogMerger` crashes on the last test (`assert($mainSuite !== null);`), because there are no log files generated to merge together. With this initial check in the WrapperRunner, we only try to merge the log files if there are actually created log files. The list with junitFiles is populated with a SplFileObject for every process/thread it will be running on, but none of the temp files are actually generated/written.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -1,4 +1,4 @@
-DOCKER_PHP_EXEC := docker compose run php
+DOCKER_PHP_EXEC := docker compose run --rm php
 
 SRCS := $(shell find ./src ./test -type f -not -path "*/tmp/*")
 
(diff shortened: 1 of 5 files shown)


Slamdunk:
Hi, can you add a test please?

CanvasCompanyHylke:
> Hi, can you add a test please?

I tried it. The single test works fine. But I can't get the whole suite working without errors locally. Hopefully it works fine in the workflow here :see_no_evil:.

Slamdunk:
Perfect, thank you

> But I can't get the whole suite working without errors locally.

That's on me, you did good :+1:

PR #870, opened 2024-08-08
Fix `baseline` reading

### LINKED ISSUE #869: Baseline configuration is not sent to PHPUnit
| Q | A
| --- | ---
| ParaTest version | 7.5.1
| PHPUnit version | 11.3.0
| PHP version | 8.3.9

#### Summary

It seems that paratest does not sends to phpunit the baseline configuration

here is an extract of my phpunit.xml:
```xml
<source ignoreSuppressionOfDeprecations="true" baseline="phpunit-baseline.xml">
</source>
```

In "normal" execution with PHPUnit, I don't see any deprecation, because they are ignored thanks to the baseline. But I see the deprecations with paratest.

At the end, I cannot configure phpunit.xml to fail on deprecations, because some of them exists, and they are not ignored with paratest.

thanks for your work!

Slamdunk:
Fix https://github.com/paratestphp/paratest/issues/869

### DIFF
--- a/composer.json
+++ b/composer.json
@@ -52,7 +52,7 @@
         "ext-pcov": "*",
         "ext-posix": "*",
         "doctrine/coding-standard": "^12.0.0",
-        "phpstan/phpstan": "^1.11.9",
+        "phpstan/phpstan": "^1.11.10",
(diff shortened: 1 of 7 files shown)


nikophil:
thanks @Slamdunk

PR #879, opened 2024-09-03
Update PHPUnit compatibility to 11.3.2

Slamdunk:


composer.json
```diff
@@ -43,9 +43,9 @@
         "phpunit/php-code-coverage": "^11.0.6",
         "phpunit/php-file-iterator": "^5.1.0",
         "phpunit/php-timer": "^7.0.1",
-        "phpunit/phpunit": "^11.3.1",
```
AJenbo, line 46: Since this just broke again with 11.3.3 should the constraint maybe be =, or < ?

Slamdunk: That would mean a release at each PHPUnit's release with the same amount of human effort in fixing non compatible release, so no thank you.
Tomorrow I'll release the fix for this one: incompatibilities are rare, I'm ok with the current release process of ParaTest

AJenbo: You wouldn't necessarily have to make an update for every single release of PHPUnit, it would be a trade off between stability and fixes.

I'm not sure how often it happens, I just saw that there where two in one week so seemed a bit frequent from just sampling that :)

Any way we sent a fix here: https://github.com/paratestphp/paratest/pull/883

Issue #868, opened 2024-08-05
Paratest encounter conflict with phar when no test path is provided


VincentLanglet:
| Q | A
| --- | ---
| ParaTest version | 7.5
| PHPUnit version | 11.3.0
| PHP version | 8.3

#### Summary

Seems like there is some differences between running
- `vendor/bin/paratest`
- `vendor/bin/paratest tests/`

#### Current behavior

I have 
- `nikic/php-parser` installed at the v5
- `phpstan/phpstan` installed (and the phar is using the v4)
- some tests using phpstan.

When I run
```
vendor/bin/paratest --runner WrapperRunner --processes 4
```

I get the error
```
Fatal error: Declaration of PhpParser\ParserAbstract::parse(string $code, ?  
  PhpParser\ErrorHandler $errorHandler = null) must be compatible with PhpPar  
  ser\Parser::parse(string $code, ?PhpParser\ErrorHandler $errorHandler = nul  
  l): ?array in phar:///var/www/wg-app/vendor/phpstan/phpstan/phpstan.phar/ve  
  ndor/nikic/php-parser/lib/PhpParser/ParserAbstract.php on line 142
```
coming from the file `tests/PHPStan/PhpDoc/TypeNodeResolverExtensionTest.php`

Which is very odd is the fact that
- I don't have the error with phpunit
- I don't have the error with `vendor/bin/paratest --runner WrapperRunner --processes 4 tests/PHPStan/PhpDoc/TypeNodeResolverExtensionTest.php`
- I don't have the error with `vendor/bin/paratest --runner WrapperRunner --processes 4 tests/`

#### How to reproduce: command, code and error stack trace

Since I have a lot of other things installed, I didn't find a small reproducer so far, but I'm trying to create one.

Still, I'd be intersted understanding why adding `tests/` change the behavior.

#### Expected behavior

No error.

Jean85:
Can you share your PHPUnit config? by adding the filter on what to launch, you're probably overriding the test suite selection coming from there,

VincentLanglet:
> Can you share your PHPUnit config? by adding the filter on what to launch, you're probably overriding the test suite selection coming from there,

Sure, I have
```xml
<?xml version="1.0" encoding="UTF-8"?>

<!-- https://phpunit.readthedocs.io/en/latest/configuration.html -->
<phpunit xmlns:xsi="http://www.w3.org/2001/XMLSchema-instance"
[...]

Issue #857, opened 2024-06-04
Repeated cli options are not working as expected


stelar7:
| Q | A
| --- | ---
| ParaTest version | 7.4.5
| PHPUnit version | 11.1.3
| PHP version | 8.2.19

#### Summary

PHPUnit 11 produces warnings when using `--exclude-group` with comma-separated values, but paratest does not work with the new syntax of passing it multiple times

See:
https://github.com/sebastianbergmann/phpunit/issues/5709 / https://github.com/sebastianbergmann/phpunit/commit/129106be139f84f2267e156025047ce5239cb8cc

#### Current behavior

`paratest --exclude-group a --exclude-group b` only passes `--exclude-group b` to PHPUnit

#### How to reproduce: command, code and error stack trace

Run with multiple `--exclude-group` options

#### Expected behavior

`paratest --exclude-group a --exclude-group b` passes all arguments to PHPUnit

PR #887, opened 2024-09-09
Keep support for PHPUnit 10

VincentLanglet:
Hi @Slamdunk, 

I don't think this PR (https://github.com/paratestphp/paratest/pull/860) require to drop the PHPUnit 10 support ; so i'd like to try working on keeping it.
Paratest don't have to support every minor/patch of PHPUnit 10, just last ones. Which means that an issue with `failOnPhpunitDeprecation` won't exist on both PHPUnit 10 and 11.

Not all project can currently, easily, using PHPUnit 11 for the following reason:
- PHPunit use nikic/php-parser 5 while other tools like PHPStan, Psalm or rector use the php-parser 4
- Symfony ErrorHandler does not work well with PHPUnit (https://github.com/symfony/symfony/issues/53812)
- I personnally have a weird issue with paratest I need to investigate more (https://github.com/paratestphp/paratest/issues/868)
- Certainly more/others.

I think it would be great for such a useful tool as paratest to keep both support of PHPUnit 10 and 11 for some time to help people migrating. I'm ok working on it. 
Currently I'm having errors locally, but since I have the same with phpunit 11 I dunno how is the state of PHPUnit 10 support with my branch, so I'd like to try on the CI.

Thanks.


Slamdunk:
PHPUnit 10 has been supported until it was afforable, then https://github.com/paratestphp/paratest/issues/857 came and it wasn't anymore.

When available, the support of two different versions of PHPUnit is merely a fortunate coincidence that as soon as it disappears we can't have it anymore.

Slamdunk:
Thanks with the support of @arnovbr, a fix has been released in [`v7.4.6`](https://github.com/paratestphp/paratest/releases/tag/v7.4.6)

PR #902, opened 2024-11-06
Require symfony/process ^6.4.14|^7.1.7

gndk:
symfony/process `<6.4.14` and `<7.1.7` contains CVE-2024-51736 which has been fixed in `6.4.14` and `7.1.7`

https://symfony.com/blog/cve-2024-51736-command-execution-hijack-on-windows-with-process-class


gndk:
Separate PRs for the 7.x and 7.4.x branches, not sure if that was the correct thing to do

PR #917, opened 2024-12-09
fix: PHPunit extensions must run before data providers

### LINKED ISSUE #916: PHPUnit `Extension::bootstrap()` called too late
| Q | A
| --- | ---
| ParaTest version | 7.6.1
| PHPUnit version | 11.4
| PHP version | 8.4

Since PHPUnit 11.4, [`bootstrap()` method is called before data provider](https://github.com/sebastianbergmann/phpunit/issues/5913), and [data provider events are dispatched synchronously](https://github.com/sebastianbergmann/phpunit/pull/5917) (before this, they were deferred).

Then, I'm wondering if this should not be reflected in paratest: otherwise, subscribers applying on data providers event will work without paratest, but will not have effect in paratest.

WDYT?

You'll find [here](https://github.com/zenstruck/foundry/pull/736) some kind of a reproducer: running test suite of this lib with paratest fails, whereas it passes with regular phpunit run. The failures occur because the extension bootstrap is called too late.

nikophil:
related to https://github.com/paratestphp/paratest/issues/916

### DIFF
--- a/composer.json
+++ b/composer.json
@@ -45,19 +45,19 @@
         "phpunit/php-timer": "^7.0.1",
         "phpunit/phpunit": "^11.5.0",
         "sebastian/environment": "^7.2.0",
-        "symfony/console": "^6.4.14 || ^7.1.7",
-        "symfony/process": "^6.4.14 || ^7.1.7"
+        "symfony/console": "^6.4.14 || ^7.2.0",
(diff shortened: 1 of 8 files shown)


nikophil:
thanks @Slamdunk

PR #929, opened 2025-02-13
Emit TestSuitLoaded event

HypeMC:
Currently extensions that rely on the TestSuitLoaded event, like the [`SymfonyExtension`](https://github.com/symfony/symfony/blob/d824d5362af59a9f697dcb368786b684987b2a03/src/Symfony/Bridge/PhpUnit/SymfonyExtension.php), might not work because the event is never dispatched in the child processes.

One example is the [`RegisterClockMockSubscriber`](https://github.com/symfony/symfony/blob/d824d5362af59a9f697dcb368786b684987b2a03/src/Symfony/Bridge/PhpUnit/Extension/RegisterClockMockSubscriber.php), which checks whether a test belongs to a certain group and registers the class to use mocked time.

### DIFF
--- a/src/WrapperRunner/ApplicationForWrapperWorker.php
+++ b/src/WrapperRunner/ApplicationForWrapperWorker.php
@@ -95,6 +95,10 @@ final class ApplicationForWrapperWorker
             $testSuite     = TestSuite::fromClassReflector($testSuiteRefl);
         }
 
+        EventFacade::emitter()->testSuiteLoaded(
+            TestSuiteBuilder::from($testSuite),
+        );
+
         (new TestSuiteFilterProcessor())->process($this->configuration, $testSuite);
 
         if ($filter !== null) {


src/WrapperRunner/ApplicationForWrapperWorker.php
```diff
@@ -95,6 +95,10 @@ public function runTest(string $testPath): int
             $testSuite     = TestSuite::fromClassReflector($testSuiteRefl);
         }
 
+        EventFacade::emitter()->testSuiteLoaded(
```
Slamdunk, line 98: According to PHPUnit's source code, this call has to be placed between this 2 calls:
https://github.com/paratestphp/paratest/blob/7771e52e192fd9bfd2c3a47139aec57a0c183bd7/src/WrapperRunner/ApplicationForWrapperWorker.php#L222-L224

Slamdunk (changes requested):

src/WrapperRunner/ApplicationForWrapperWorker.php
HypeMC: The call requires `$testSuite`, so this would mean that `bootstrap()` would need to be called after `TestSuite::fromClassReflector()`, but `TestSuite::fromClassReflector()` requires the configuration to be bootstrapped. Not sure what the best option here is.

Slamdunk: I see.

We indeed need the bootstrap to be separated from the test suite loading phase, because in contrary of PHPUnit we load multiple suites per bootstrap.

Let's try your fix and see if we break someone's build :wink:

Slamdunk (approved):

PR #945, opened 2025-04-30
feat: pass `--enforce-time-limit` and `--default-time-limit` to PHPUnit worker

### LINKED ISSUE #944: Add support for --enforce-time-limit and --default-time-limit
Hey,

I noticed that it's not possible to provide --enforce-time-limit or --default-time-limit <sec> into paratest

```
The "--enforce-time-limit" option does not exist.
```

But setting these within the phpunit.xml work fine, It would be nice if these were possible to provide via the CLI as well 

Thanks

ALameLlama:
resolves: https://github.com/paratestphp/paratest/issues/944


ALameLlama:
let me know if any changes are needed :)

src/Options.php
```diff
@@ -571,6 +573,19 @@ public static function setInputDefinition(InputDefinition $inputDefinition): voi
                 InputOption::VALUE_NONE,
                 '@see PHPUnit guide, chapter: ' . $chapter,
             ),
+            new InputOption(
+                'enforce-time-limit',
+                null,
+                InputOption::VALUE_NONE,
+                '@see PHPUnit guide, chapter: ' . $chapter,
```
Slamdunk: The chapter is wrong: please move it between `disallow-test-output` and `dont-report-useless-tests`

Slamdunk (changes requested):
Please add a test with a `sleep(2);`, enforce with flag the time limit to be `1` and expect the test to fail

Slamdunk (approved):

ALameLlama:
> Please add a test with a `sleep(2);`, enforce with flag the time limit to be `1` and expect the test to fail

I've added a test in a new PR :)

Issue #695, opened 2022-09-15
Exception when running paratest_for_phpstorm with --filter option 


slapoguzov:
| Q | A
| --- | ---
| ParaTest version | 6.6.3
| PHPUnit version | 9.5.24
| PHP version | 8.1

#### Summary

When I try to run `paratest_for_phpstorm` with `--filter` option, for example:
```
/opt/homebrew/bin/php -dxdebug.mode=debug -dxdebug.client_port=9000 -dxdebug.client_host=127.0.0.1 /Users/Aleksandr.Slapoguzov/PhpstormProjects/paratest_filter/vendor/bin/paratest_for_phpstorm /Users/Aleksandr.Slapoguzov/PhpstormProjects/paratest_filter/vendor/phpunit/phpunit/phpunit --bootstrap /private/var/folders/nq/fh_373lj7p3fjqq5jvs5rg6c0000kt/T/ide-phpstorm_bootstrap.php --filter "/MyTest::testMe( .*)?$$/" /Users/Aleksandr.Slapoguzov/PhpstormProjects/paratest_filter/tests --teamcity
```

the following exception is thrown:

```
/opt/homebrew/bin/php -dxdebug.mode=debug -dxdebug.client_port=9000 -dxdebug.client_host=127.0.0.1 /Users/Aleksandr.Slapoguzov/PhpstormProjects/paratest_filter/vendor/bin/paratest_for_phpstorm /Users/Aleksandr.Slapoguzov/PhpstormProjects/paratest_filter/vendor/phpunit/phpunit/phpunit --bootstrap /private/var/folders/nq/fh_373lj7p3fjqq5jvs5rg6c0000kt/T/ide-phpstorm_bootstrap.php --filter "/MyTest::testMe( .*)?$$/" /Users/Aleksandr.Slapoguzov/PhpstormProjects/paratest_filter/tests --teamcity
Testing started at 12:12 PM ...
Testing started at 12:12 PM ...
PHP Fatal error:  Uncaught RuntimeException: Missing path to 'vendor/brianium/paratest/bin/paratest' in /Users/Aleksandr.Slapoguzov/PhpstormProjects/paratest_filter/vendor/brianium/paratest/src/Util/PhpstormHelper.php:63

Stack trace:
[...]

PR #692, opened 2022-09-08
Fixes for phpstormhelper to do with re-running failed tests

billypoke:
I have been getting console messages caused by the previous code I contributed when using the "re-run failed tests" button. This PR resolves those errors. I would really appreciate another set of eyes on this as well, as some of the removed code didn't seem to serve a purpose and I wanted to make sure there wasn't a use-case I was missing.

Also re-write for the util test to have more documentation and better-explained test cases

PR #995, opened 2025-08-15
Fix Incompatibilty With PhpStorm on Windows

FeBe95:
This is a bug fix for the following error:
```sh
Fatal error: Uncaught RuntimeException
Missing path to '/paratest_for_phpstorm' in C:\repos\some_repo\vendor\brianium\paratest\src\Util\PhpstormHelper.php:60
```

## Detailed Description
This error occurred when trying to use the "Rerun Failed Test" functionality in PhpStorm on a Windows machine. This only happened if the run configuration option "Use ParaTest" is set by the user and contains backward slashes. On Windows this is very common, since the native Windows Explorer "Select Path" dialog can be used.

grafik

## Root Cause
When this option is set, the resulting command line contains mixed forward and backward slashes, because PhpStorm inserts the path as is:
```sh
C:\php\php.exe C:\repos\some_project\vendor\bin\paratest_for_phpstorm C:/repos/some_project/vendor/phpunit/phpunit/phpunit --configuration C:\repos\some_project\phpunit.xml --filter "..." C:\xampp\htdocs\bidX\tests\Unit --teamcity
```

Therefore, calling
```php
self::getArgvKeyFor($argv, '/paratest_for_phpstorm')
```
throws the previously mentioned error, because it tries to find the `paratest_for_phpstorm` binary prefixed with a forward slash.

## The Fix
Search for both `/paratest_for_phpstorm` and `\paratest_for_phpstorm` in the provided arguments.

## References
This PR relates to these issues:
- #692
- #695

### DIFF
--- a/src/Util/PhpstormHelper.php
+++ b/src/Util/PhpstormHelper.php
@@ -10,6 +10,7 @@ use function array_search;
 use function array_unshift;
 use function in_array;
 use function str_ends_with;
+use function str_replace;
 
 /** @internal */
 final readonly class PhpstormHelper
@@ -48,7 +49,7 @@ final readonly class PhpstormHelper
     /** @param  array<int, string> $argv */
     private static function getArgvKeyFor(array $argv, string $searchFor): int
     {
-        $searchForForwardSlash = str_replace('\\', '/', $searchFor);
+        $searchForForwardSlash  = str_replace('\\', '/', $searchFor);
         $searchForBackwardSlash = str_replace('/', '\\', $searchFor);
 
         foreach ($argv as $key => $arg) {
--- a/src/Util/PhpstormHelper.php
+++ b/src/Util/PhpstormHelper.php
@@ -48,8 +48,11 @@ final readonly class PhpstormHelper
     /** @param  array<int, string> $argv */
     private static function getArgvKeyFor(array $argv, string $searchFor): int
     {
+        $searchForForwardSlash = str_replace('\\', '/', $searchFor);
+        $searchForBackwardSlash = str_replace('/', '\\', $searchFor);
+
         foreach ($argv as $key => $arg) {
-            if (str_ends_with($arg, $searchFor)) {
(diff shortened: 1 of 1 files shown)


src/Util/PhpstormHelper.php
```diff
@@ -48,8 +48,11 @@ public static function handleArgvFromPhpstorm(array &$argv, string $paratestBina
     /** @param  array<int, string> $argv */
     private static function getArgvKeyFor(array $argv, string $searchFor): int
     {
+        $searchForForwardSlash = str_replace('\\', '/', $searchFor);
+        $searchForBackwardSlash = str_replace('/', '\\', $searchFor);
```
Slamdunk:
```suggestion
        $searchFor = str_replace('/', DIRECTORY_SEPARATOR, $searchFor);
```
Instead of a double search, what about a single search with the os-aware `DIRECTORY_SEPARATOR` constant?

Slamdunk (changes requested):
Instead of a double search, what about a single replace

src/Util/PhpstormHelper.php
FeBe95: Unfortunately, both forward and backward slashes can be used within one single command line on Windows, as you can see in the example I provided above. The command line contains:
- `C:\repos\some_project\vendor\bin\paratest_for_phpstorm` and
- `C:/repos/some_project/vendor/phpunit/phpunit/phpunit`. 

The `getArgvKeyFor()` method must support both types. Therefore, it cannot rely on `DIRECTORY_SEPARATOR` as a replacement in `$searchFor` because `$argv` might contain non-OS-standard slashes. Users have the "freedom of choice" to use any type of slash when providing the configuration path.

P.S. Why did I use `str_replace()` twice? On Windows, a single path can contain both types of slashes at the same time and still be valid, e.g., `C:\repos\some_project\vendor\bin/paratest_for_phpstorm`. Using `str_replace()` twice normalizes all paths to contain only one type of slash for easier comparison.

Slamdunk (approved):

Slamdunk:
Please fix the build so we can merge and ship it :muscle:

PR #1013, opened 2025-09-09
feat: add `--shard` option

babeuloula:
This pull request introduces a new `--shards` option that allows sharding the test suite into multiple parts.

The main use case is improving CI performance in scenarios where running the full suite in parallel processes still takes too long. By dividing the suite into shards, it becomes possible to run multiple jobs in parallel, each executing only a portion of the tests.

# Usage

The option follows the format --shards X/Y, where:
- Y is the total number of shards to divide the suite into
- X is the index of the shard to run (starting at 1)

## Example:

- `--shards 1/2` runs the first half of the test suite
- `--shards 2/2` runs the second half

# Motivation

In CI environments, test startup and execution time can remain high even with parallel processes. This feature enables distributing the suite across multiple jobs, reducing overall CI runtime and improving feedback speed.

### DIFF
--- a/src/Options.php
+++ b/src/Options.php
@@ -6,6 +6,7 @@ namespace ParaTest;
 
 use Fidry\CpuCoreCounter\CpuCoreCounter;
 use Fidry\CpuCoreCounter\NumberOfCpuCoreNotFound;
+use InvalidArgumentException;
 use PHPUnit\TextUI\Configuration\Builder;
 use PHPUnit\TextUI\Configuration\Configuration;
 use RuntimeException;
@@ -28,6 +29,7 @@ use function is_array;
 use function is_bool;
 use function is_numeric;
 use function is_string;
+use function preg_match;
 use function realpath;
 use function sprintf;
 use function str_starts_with;
@@ -103,6 +105,8 @@ final readonly class Options
         public string $tmpDir,
         public bool $verbose,
         public bool $functional,
+        public int $currentShard,
+        public int $totalShards,
     ) {
         $this->needsTeamcity = $configuration->outputIsTeamCity() || $configuration->hasLogfileTeamcity();
     }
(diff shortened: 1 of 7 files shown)


src/Options.php
```diff
@@ -275,6 +296,12 @@ public static function setInputDefinition(InputDefinition $inputDefinition): voi
                 InputOption::VALUE_NONE,
                 'Output more verbose information',
             ),
+            new InputOption(
+                'slices',
```
Slamdunk: The term I see in other test frameworks and languages for this very feature is `shard`, see for example https://playwright.dev/docs/test-sharding

Please reflect the needed change in all the variables and methods as well

src/Options.php
```diff
@@ -157,6 +160,22 @@ public static function fromConsoleInput(InputInterface $input, string $cwd): sel
...
 
+        assert(is_string($options['slices'] ?? ''));
+        $parts        = [];
+        $currentSlice = $totalSlices = 0;
+        $pregMatch    = preg_match('/^([0-9]+)\/([0-9]+)$/', $options['slices'] ?? '', $parts);
+        if ($pregMatch === 1) {
+            $currentSlice = (int) $parts[1];
+            $totalSlices  = (int) $parts[2];
+
+            // Validate slice parameters - if invalid, reset to 0 (no slicing)
```
Slamdunk: We don't need to validate input, just parse it: if `$options['slices']` is not null and doesn't exacty match what we need, throw an Exception

src/WrapperRunner/SuiteLoader.php
```diff
@@ -214,4 +222,36 @@ private function stripCwd(string $filename): string
...
+    {
+        $tests = $this->extractTestsInSuite($suite);
+
+        $slices        = $this->options->totalSlices;
+        $current       = $this->options->currentSlice - 1; // 0 indexed. Slice 1 is in reality slice 0
+        $total         = count($tests);
+        $testsPerSlice = (int) ceil($total / $slices);
+        $offset        = $testsPerSlice * $current;
+
+        $suite->setTests(
```
Slamdunk: I am not entirely sure we can do this safely, the `TestSuite` has states regarding groups, cache and so on.

I'd feel more safe instantiating a new `TestSuite` instance

test/Unit/WrapperRunner/ResultPrinterTest.php
```diff
@@ -118,6 +118,36 @@ public function testStartPrintsOptionInfoWithSingularForOneProcess(): void
...
+    {
+        $this->printer = new ResultPrinter($this->output, $this->createOptionsFromArgv([
+            '--slices' => 'invalid',
+            '--verbose' => true,
+        ]));
+        $contents      = $this->getStartOutput();
+
+        self::assertStringNotContainsString('Slices:', $contents);
+    }
+
```
Slamdunk, line 139: These tests are ok, but I'd like to be more thorough: please add a new test in the `WrapperRunnerTest` as well, with 2 shards, and do extended assertions on output and jUnit logs

Slamdunk (changes requested):
Thank you very much for this :muscle: 

I bet that after we release this, someone will also ask a way to merge back code coverage and logs, but I'll ask them do code it :stuck_out_tongue_closed_eyes:

Slamdunk (approved):

PR #1028, opened 2025-09-25
TestDox: add `--testdox-text` and `--testdox-html` support

### LINKED ISSUE #1005: Support --testdox-text and --testdox-html options
PHPUnit has the `--testdox-text` and `--testdox-html` options (https://docs.phpunit.de/en/10.5/textui.html#logging), which log testdox output to a file. This option is currently not supported by paratest. As it seems that paratest already includes logic to merge testdox results, I hope that it would not be much effort to support these two logging options as well.

aboks:
Fixes #1005

### DIFF
--- a/src/TestDox/TestDoxResultsMerger.php
+++ b/src/TestDox/TestDoxResultsMerger.php
@@ -1,0 +1,125 @@
+<?php
+
+declare(strict_types=1);
+
+namespace ParaTest\TestDox;
+
+use PHPUnit\Logging\TestDox\TestResult as TestDoxTestMethod;
+use PHPUnit\Logging\TestDox\TestResultCollection;
+use PHPUnit\Logging\TestDox\TestResultCollection as TestdoxTestResultCollection;
+use ReflectionException;
+use ReflectionMethod;
+use SplFileInfo;
+
+use function array_merge;
+use function assert;
+use function file_get_contents;
+use function is_subclass_of;
+use function ksort;
+use function uksort;
+use function unserialize;
+use function usort;
+
+/** @internal */
+final readonly class TestDoxResultsMerger
+{
+    /**
+     * @param list<SplFileInfo> $testdoxFiles
+     *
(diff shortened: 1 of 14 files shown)


Slamdunk:
Thanks, I'll have a look at this in the next weeks

aboks:
Thanks. Supporting this was a bit more work than I thought. I had to rework the existing testdox support to not write the plain text testdox output to the per-worker testdox files, but rather the serialized collected testdox data structure, somewhat similar to how this is done for the 'normal' test result data. In the main process, these are then unserialized, merged and printed to the desired outputs (console and/or logfiles).

src/TestDox/TestDoxResultsMerger.php
```diff
@@ -0,0 +1,125 @@
...
+     * @param list<SplFileInfo> $testdoxFiles
+     *
+     * @return array<string,TestdoxTestResultCollection>
+     */
+    public function getResultsFromTestdoxFiles(array $testdoxFiles): array
+    {
+        /** @var array<string,TestdoxTestResultCollection> $testMethodsGroupedByClass */
+        $testMethodsGroupedByClass = [];
+        foreach ($testdoxFiles as $testdoxFile) {
+            $testdoxFileContents = file_get_contents($testdoxFile->getPathname());
```
Slamdunk, line 39:
```suggestion
            if (! $testdoxFile->isFile()) {
                continue;
            }
            $testdoxFileContents = file_get_contents($testdoxFile->getPathname());
```
This assumption is wrong: if the number of tests is lower than the available processes, the Nth file is never written so it doesn't exists.

Slamdunk (changes requested):
Most excellent work :pinched_fingers: 

Just one fix and we can ship this

aboks:
Ah thanks, I didn't know that. I added the fix to my branch.

Slamdunk (approved):

Issue #1044, opened 2025-11-05
slow coverage data generation


staabm:
| Q | A
| --- | ---
| ParaTest version | 7.8.4
| PHPUnit version | 11.5.43
| PHP version | 8.3.27

#### Summary

running paratest with pcov and coverage generation on phpstan-src reports, that after running the tests for 4min40s (which is fast), we need another minute to generate the coverage xml report

<img width="1384" height="378" alt="Image" src="https://github.com/user-attachments/assets/0e2b43fc-6be8-44d4-a768-c93e65ddfba4" />

see https://github.com/phpstan/phpstan-src/actions/runs/19078182639/job/54499398288?pr=4525

I would hope that there is some low hanging fruit to speedup coverage xml export.
I did not yet succeed in extracting the gathered information to make the process easily reproducible (and measurable with a profiler).
maybe you have ideas on how this can be made faster or how to at best investigate the problem without a 8minute feedback loop?

#### Current behavior

coverage data generation seems slow

#### How to reproduce: command, code and error stack trace

- git clone phpstan-src
- `composer install`
- run `make tests` (which installs paratest)
- run `php -d pcov.enabled=1 tests/vendor/bin/paratest --passthru-php="'-d' 'pcov.enabled=1'" --coverage-xml=tmp/coverage/coverage-xml --log-junit=tmp/coverage/junit.xml`

#### Expected behavior

faster xml code coverage export

Slamdunk:
The complete absence of `#[CoversClass(***)]` annotations for sure kills performance.
Without them, PHPUnit needs to track every source code for every test, multiple that by the number of processes, and merge everything at the end.

Just to check if that's the culprit, code this *silly* try:

1. for every `FooTest.php` file in the `tests/` folder, add a single `#[CoversClass(Foo::class)]` annotation if `Foo` exists
2. otherwhise add a `#[CoversNothing]` one

If the drop in the test duration is much higher than the drop in the overall code coverage, you've found the missing piece of the puzzle

Jean85:
Would it be possible to save some time by merging the coverage information "on the fly"? In Paraunit I do that incrementally as soon as a single process terminates its tests.

Slamdunk:
@Jean85 I don't think that would make any meaninful difference, but I'd be happy to be proven wrong

staabm:
I just [added `CoversNothing` to 13 integration test-classes](https://github.com/phpstan/phpstan-src/pull/4535) which had interessintg results:

[...]

PR #1045, opened 2025-11-10
Incrementally create coverage report

staabm:
closes https://github.com/paratestphp/paratest/issues/1044

using 7.8.x as target branch, because thats what PHPStan is using (because of compat with PHP 8.2)

### DIFF
--- a/src/Coverage/CoverageMerger.php
+++ b/src/Coverage/CoverageMerger.php
@@ -17,10 +17,10 @@ final readonly class CoverageMerger
     ) {
     }
 
-    public function addCoverageFromFile(SplFileInfo $coverageFile): void
+    public function addCoverageFromFile(SplFileInfo $coverageFile): bool
     {
         if (! $coverageFile->isFile() || $coverageFile->getSize() === 0) {
-            return;
+            return false;
         }
 
         /** @psalm-suppress UnresolvableInclude **/
@@ -28,5 +28,7 @@ final readonly class CoverageMerger
         assert($coverage instanceof CodeCoverage);
 
         $this->coverage->merge($coverage);
+
+        return true;
     }
 }
--- a/src/WrapperRunner/WrapperRunner.php
+++ b/src/WrapperRunner/WrapperRunner.php
@@ -369,12 +369,16 @@ final class WrapperRunner implements RunnerInterface
 
     protected function finalizeCodeCoverageReports(): void
     {
-        if ($this->coverageManager === null) {
+        if ($this->coverageFiles === []) {
             return;
         }
 
         while (count($this->coverageFiles) > 0) {
             $this->generateIncrementalCodeCoverageReport();
+        }
+
+        if ($this->coverageManager === null) {
+            return;
         }
 
         $this->coverageManager->generateReports(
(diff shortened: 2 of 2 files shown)


staabm:
in parallel I have also improved the PHPUnit XML generation, so it is twice as fast
- https://github.com/theseer/tokenizer/pull/25
- https://github.com/theseer/tokenizer/pull/29
- https://github.com/theseer/tokenizer/pull/24
- https://github.com/theseer/tokenizer/pull/32
- https://github.com/theseer/tokenizer/pull/35

staabm:
@Slamdunk you might consider cherry picking [58ef344](https://github.com/paratestphp/paratest/pull/1045/commits/58ef344e569d015a10dd70c9ddd7d2af048ee751) as it is unrelated to this PR, but required to fix PHPStan errors

regarding tests: I cannot get a green build for unit tests locally on the 7.8.x branch even before this PR, therefore I am not sure, whether the test errors are related to the PR or not

staabm:
ok, have a working copy now. need to look into the errors

staabm:
thinking more about this PR.. I think this approach would only make sense when worker processes would stop/start from time to time while running tests. e.g. a worker process can end after it executed X batches of tests and a new one would be spawned (this would also help against memory leaks in tests).

otherwise the coverage.xml files of the worker would not be "finished" and cannot be processed early.

As is, my understanding is, that most/all workers are alive until all tests are processed, therefore the coverage merging seems to not save much time as it does not really happen in parallel

staabm:
> As is, my understanding is, that most/all workers are alive until all tests are processed, therefore the coverage merging seems to not save much time as it does not really happen in parallel

while testing more I realized that worker processes at least start/stop when process-isolation is enabled

Jean85:
> As is, my understanding is, that most/all workers are alive until all tests are processed, therefore the coverage merging seems to not save much time as it does not really happen in parallel

If this is true, this would be profoundly different from what I have achieved in Paraunit. There, each test is a single process, so I do the merging every time that a process ends: https://github.com/facile-it/paraunit/blob/189def486646c47392f4ecf6055c8ffb297ea1ab/src/Coverage/CoverageMerger.php#L28-L47

Once all processes are done, I then convert the coverage result from a PHP object to the desired format: https://github.com/facile-it/paraunit/blob/2.x/src/Coverage/CoverageResult.php

staabm:
> thinking more about this PR.. I think this approach would only make sense when worker processes would stop/start from time to time while running tests. e.g. a worker process can end after it executed X batches of tests and a new one would be spawned (this would also help against memory leaks in tests).

just found this code, which does what I am talking about: https://github.com/paratestphp/paratest/blob/de774a70f6ec846b9fed0361cd9af32ee71ab49e/src/WrapperRunner/WrapperRunner.php#L153-L156



> If this is true

I am still in the process of wrapping my head arround paratest.. please take my observations with a grain of salt

staabm:
I think the fundamental flaw in this PR is, that coverage-merging itself is not slow, but the report generation with `coverageManager->generateReports`, which still happens in a single step at the very end within `finalizeCodeCoverageReports`.

PR #1054, opened 2025-11-29
Add support for Symfony 8, drop Symfony 6

ngrie:
Adds support for Symfony 8 while still supporting the previous versions.

Closes #1053
Closes #1052
Closes #1051

### DIFF
--- a/composer.json
+++ b/composer.json
@@ -45,8 +45,8 @@
         "phpunit/php-timer": "^8",
         "phpunit/phpunit": "^12.4.4",
         "sebastian/environment": "^8.0.3",
-        "symfony/console": "^6.4.20 || ^7.3.4",
-        "symfony/process": "^6.4.20 || ^7.3.4"
+        "symfony/console": "^7.3.4 || ^8.0.0",
+        "symfony/process": "^7.3.4 || ^8.0.0"
     },
     "require-dev": {
         "ext-pcntl": "*",
@@ -57,7 +57,7 @@
         "phpstan/phpstan-deprecation-rules": "^2.0.3",
         "phpstan/phpstan-phpunit": "^2.0.8",
         "phpstan/phpstan-strict-rules": "^2.0.7",
-        "symfony/filesystem": "^6.4.13 || ^7.3.2"
+        "symfony/filesystem": "^7.3.2 || ^8.0.0"
     },
     "autoload": {
         "psr-4": {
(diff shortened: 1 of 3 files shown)


.github/workflows/ci.yml
```diff
@@ -17,7 +17,7 @@ jobs:
     strategy:
       matrix:
         php-version:
-          - "8.3"
+          - "8.4"
```
Slamdunk: Please let's keep the lower supported PHP version as the main one for the build

composer.json
```diff
@@ -45,8 +45,8 @@
         "phpunit/php-timer": "^8",
         "phpunit/phpunit": "^12.4.4",
         "sebastian/environment": "^8.0.3",
-        "symfony/console": "^6.4.20 || ^7.3.4",
-        "symfony/process": "^6.4.20 || ^7.3.4"
+        "symfony/console": "^6.4.20 || ^7.3.4 || ^8.0.0",
+        "symfony/process": "^6.4.20 || ^7.3.4 || ^8.0.0"
```
Slamdunk: We can drop `symfony:v6`: two major versions are enough for a free software

Slamdunk (changes requested):
Thank you for taking care of this. A few tweaks and we can ship it

.github/workflows/ci.yml
```diff
@@ -147,7 +147,7 @@ jobs:
...
```
ngrie: @Slamdunk Thanks four your review!

I reverted the other changes in this file but kept this one. Since Symfony 8 requires PHP 8.4, running this job using PHP 8.3 would result in Symfony 7.4 components being installed, and thus a different output that is generated by PHPStan.

Do you want me to revert this line as well anyways?

Slamdunk: Yes please: all the tools must run on the lowest supported version, and only the tests against every version

Slamdunk (approved):

composer.json
gnutix: If there was no technical reason to drop it, it's too bad to do so. It's still actively supported until November and 1 more year for security fixes. Meaning it's totally fine to use. Upgrading to 7.x has tons of BC breaks, while upgrading to the next minor version of paratest doesn't. It just means your users will lose your library's bug fixes and improvements for a year or two.

PR #1056, opened 2025-12-06
Add `--exclude-source-from-xml-coverage` support

staabm:
tested it manually with 
(since the test-suite seem not to have a test for coverage-xml).

```
XDEBUG_MODE=coverage bin/paratest --coverage-xml tmp/coverage 
...
Generating code coverage report in PHPUnit XML format ... done [00:00.097]
```
and
```
XDEBUG_MODE=coverage bin/paratest --coverage-xml tmp/coverage --exclude-source-from-xml-coverage
...
Generating code coverage report in PHPUnit XML format ... done [00:00.046]
```
note: the report generation is 50% faster

I verified manually that the generated report no longer contains a `<source>` element, when invoked with `--exclude-source-from-xml-coverage`

refs https://github.com/sebastianbergmann/phpunit/issues/6422


Slamdunk:
Would it make sense to add it to `\ParaTest\Options::OPTIONS_TO_KEEP_FOR_PHPUNIT_IN_WORKER` as well?

staabm:
> Would it make sense to add it to `\ParaTest\Options::OPTIONS_TO_KEEP_FOR_PHPUNIT_IN_WORKER` as well?

tbh, I am not sure.

there is another similar option which refines `coverage-text` - namely `only-summary-for-coverage-text` - which is not contained in `OPTIONS_TO_KEEP_FOR_PHPUNIT_IN_WORKER` either.

Slamdunk:
> there is another similar option[...]

Don't take for granted that every option has been thoroughly scrutinized in the making of `OPTIONS_TO_KEEP_FOR_PHPUNIT_IN_WORKER` :sweat_smile: 

Give it a try

staabm:
> Give it a try

I added it and it does not seem to make a difference from what I can tell.
(I don't know where to look for though)

staabm:
don't know what todo with this PHPStan error, as I am not able to reproduce it locally

staabm:
@Slamdunk I think it should be good to go

Slamdunk (approved):

PR #1062, opened 2026-01-06
Fix: Detect and fail when workers crash after test execution but before writing result files

gpn273:
#  Problem

When a child process crashes unexpectedly (e.g., due to out-of-memory errors) after executing tests but before writing result/coverage files, Paratest would silently ignore the missing files and exit with code 0. This causes false-positive CI pipeline passes despite test failures.

## Scenario

1. Test executes and reports error (E) in output
2. Status file is written (worker marked as "free")
3. Process crashes due to OOM before `application->end()` completes
4. Result/coverage files are never written
5. Paratest exits successfully (exit code 0)

# Solution

Added validation to ensure all workers that executed tests have written their result and coverage files. If files are missing, throw `MissingResultsException` with a clear error message.

## Changes

- Added `WrapperWorker::hasExecutedTests()` to track which workers ran tests
- Added `WrapperRunner::$requiredTestResultFiles` and `$requiredCoverageFiles` to track file paths that must exist
- Validate test result files only for workers that executed tests (not idle workers)
- Validate coverage files only for workers that executed tests and have coverage enabled
- Throw `MissingResultsException` with actionable error message when files are missing

## Testing

- New `MissingResultsException` unit tests verify exception behaviour
- Validated fixes in a code base which replicates the issue which this pull request solves

## Fixes

Addresses the issue where Paratest would pass CI despite worker crashes when using `--passthru-php='-dmemory_limit=1G'` or similar memory constraints.

### Screenshot

Screenshot demonstrating the exception raising when test result files are missing for a given worker:

image

### DIFF
--- a/src/WrapperRunner/MissingResultsException.php
+++ b/src/WrapperRunner/MissingResultsException.php
@@ -1,0 +1,33 @@
+<?php
+
+declare(strict_types=1);
+
+namespace ParaTest\WrapperRunner;
+
+use RuntimeException;
+
+use function implode;
+use function sprintf;
+
+/** @internal */
+final class MissingResultsException extends RuntimeException
+{
+    /**
+     * @param list<non-empty-string>   $missingFiles
+     * @param 'test_result'|'coverage' $fileType
+     */
+    public static function create(array $missingFiles, string $fileType): self
+    {
+        $fileTypeLabel = $fileType === 'test_result' ? 'test result' : 'coverage';
+
+        $message = sprintf(
+            'One or more workers failed to generate %s files, likely due to unexpected process termination (e.g., out of memory). ' .
+            'Missing %s files: %s',
+            $fileTypeLabel,
+            $fileTypeLabel,
+            implode(', ', $missingFiles),
+        );
+
+        return new self($message);
+    }
+}
--- a/src/WrapperRunner/WrapperRunner.php
+++ b/src/WrapperRunner/WrapperRunner.php
@@ -32,9 +32,10 @@ use function assert;
 use function count;
 use function dirname;
 use function file_get_contents;
+use function filesize;
+use function is_file;
 use function max;
 use function realpath;
-use function spl_object_id;
 use function unlink;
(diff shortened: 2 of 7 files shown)


Slamdunk:
This is a welcome contribution, thank you for the explanation and the fix.

I'm wondering: can we also add a test that reproduce the issue?

An idea could be to create a test that registers a shutdown function to delete the result file and the coverage file, they can be easeily gathered within the test by copy-pasting the `getopt` line of `bin/phpunit-wrapper.php`: would you be so kind to give it a try please?

Jean85:
> An idea could be to create a test that registers a shutdown function to delete the result file and the coverage file, they can be easeily gathered within the test by copy-pasting the `getopt` line of `bin/phpunit-wrapper.php`: would you be so kind to give it a try please?

Stealing from my own ideas, you could also:
 * force a fatal error with an anonymous class that doesn't implement all the right methods: https://github.com/facile-it/paraunit/blob/c5bdee0de7c5292df1a0377fe6f6869f9c318177/tests/Stub/FatalErrorTestStub.php#L11
 * use a simple `exit()` (`139` is the exit code for segfaults): https://github.com/facile-it/paraunit/blob/c5bdee0de7c5292df1a0377fe6f6869f9c318177/tests/Stub/SegFaultTestStub.php#L11

gpn273:
Thank you both (@Slamdunk & @Jean85) for your feedback.

I've added test cases that reproduce the issue using the approach you suggested.

The test fixtures register shutdown functions that find file paths from `$_SERVER['argv']` (for test result files) and unserialize `--phpunit-argv` (for coverage files), wait for the files to be written, then delete them to simulate an incomplete `application->end()` call. I've added three test cases covering: missing test result files, missing both files, and missing only coverage files.

All 128 tests now pass, and the tests verify that `MissingResultsException` is thrown with appropriate error messages for each scenario.

Slamdunk (approved):
Thank you for this very well written PR 💪

PR #1066, opened 2026-02-04
Add `--max-processes` option

HypeMC:
This PR adds a `--max-processes` option that, when used with `--processes=auto`, limits the number of processes so it does not exceed a specified maximum. This is useful when the number of CPU cores on CI servers varies, but you don't want to use all of them in some cases.

### DIFF
--- a/test/Unit/OptionsTest.php
+++ b/test/Unit/OptionsTest.php
@@ -49,6 +49,13 @@ final class OptionsTest extends TestBase
         self::assertEquals(Options::getNumberOfCPUCores(), $options->processes);
     }
 
+    public function testAutoProcessesModeWithMaxProcesses(): void
+    {
+        $options = $this->createOptionsFromArgv(['--processes' => 'auto', '--max-processes' => '0']);
+
+        self::assertSame(0, $options->processes);
+    }
+
     public function testPassthru(): void
     {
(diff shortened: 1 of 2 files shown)


test/Unit/OptionsTest.php
```diff
@@ -49,6 +50,13 @@ public function testAutoProcessesMode(): void
         self::assertEquals(Options::getNumberOfCPUCores(), $options->processes);
     }
 
+    public function testAutoProcessesModeWithMaxProcesses(): void
+    {
+        $options = $this->createOptionsFromArgv(['--processes' => 'auto', '--max-processes' => '3']);
+
+        self::assertEquals(min(Options::getNumberOfCPUCores(), 3), $options->processes);
```
Slamdunk:
```suggestion
        $options = $this->createOptionsFromArgv(['--processes' => 'auto', '--max-processes' => '0']);

        self::assertSame(0, $options->processes);
```

Copy-pasting part of the `src/` code defeats the purpose of a test. We can have a more effective test by pushing the boundaries on non-so meaningfull cases

Slamdunk (changes requested):

test/Unit/OptionsTest.php
HypeMC: Done

Slamdunk (approved):

PR #1074, opened 2026-02-06
Add support for `--display-all-issues`

### LINKED ISSUE #1073: Add support for --display-all-issues
In [PHPUnit 12.1.4](https://github.com/sebastianbergmann/phpunit/releases/tag/12.1.4), specifically commit https://github.com/sebastianbergmann/phpunit/commit/3a0a6308a4e8acf6e028efcaa66639e58c304959, the `--display-all-issues` / `displayDetailsOnAllIssues="true"` option was added ([documentation](https://docs.phpunit.de/en/12.1/configuration.html#the-displaydetailsonallissues-attribute)). In short: It makes PHPUnit show messages for *all* issues, instead of having to set the various options for specific issue types separately, and will include future issue types as well. Very handy.

This option currently doesn't work in Paratest: It cannot be passed via the command line, and is not used when set in the XML configuration.

Reason it doesn't work with the XML configuration despite using PHPUnit's own code for parsing the XML configuration is that Paratest constructs the `DefaultResultPrinter` itself ([here](https://github.com/paratestphp/paratest/blob/v7.18.0/src/WrapperRunner/ResultPrinter.php#L182-L255)), and that is the place where PHPUnit itself applies this new option ([here](https://github.com/sebastianbergmann/phpunit/blob/13.0.0/src/TextUI/Output/Facade.php#L196-L254)).

Slamdunk:
Fix https://github.com/paratestphp/paratest/issues/1073


DvdGiessen:
Hah, that was fast! 🎉 Thank you so much!

Can confirm v7.19.0 now correctly prints issues using the `displayDetailsOnAllIssues` option.

PR #1088, opened 2026-03-10
Add support for `php-code-coverage:v14`, require `phpunit/phpunit:v13.1`

Slamdunk:


staabm:
I am really curious whether the use of the new php-code-coverage lib will speedup coverage generation/merge

composer.json
```diff
@@ -33,20 +33,20 @@
...
-        "phpunit/phpunit": "^12.5.14 || ^13.0.5",
-        "sebastian/environment": "^8.0.3 || ^9",
-        "symfony/console": "^7.4.7 || ^8.0.7",
-        "symfony/process": "^7.4.5 || ^8.0.5"
+        "phpunit/php-code-coverage": "^14",
+        "phpunit/php-file-iterator": "^7",
+        "phpunit/php-timer": "^9",
+        "phpunit/phpunit": "^13.1",
+        "sebastian/environment": "^9.1",
+        "symfony/console": "^7.4.8 || ^8.1.0",
```
jnoordsij, line 48: I just tried updating to the latest version (including a bump to PHPUnit 13.1), however this constraint prevents me from doing so. Is there any particular reason to require the ([unreleased](https://symfony.com/releases/8.1)) 8.1 minor here as minimum, or should this have been the latest 8.0.x patch instead?

Slamdunk: Whops, my bad sorry, fix released in [`v7.21.1`](https://github.com/paratestphp/paratest/releases/tag/v7.21.1)

jnoordsij: Ah no worries, thanks for the fix!

timnolte:
Why did this change drop support for PHP 8.3, this was not documented in any of the releases either. Shouldn't dropping support for a version of PHP be considered a breaking change that results in a major point release. It makes it difficult and incorrect version locking to be doing this in a Minor release.

Slamdunk:
> Why did this change drop support for PHP 8.3

Because PHP 8.4 is required by the very dependency it's upgraded here

> Shouldn't dropping support for a version of PHP be considered a breaking change that results in a major point release

A BC break happens when your application breaks with a new release of a dependency.
This isn't the case, so it can land in a minor release.

> It makes it difficult and incorrect version locking to be doing this in a Minor release.

Quite the contrary: this way you can upgrade your PHP version knowing nothing breaks

timnolte:
> > Why did this change drop support for PHP 8.3
> 
> Because PHP 8.4 is required by the very dependency it's upgraded here
> 
> > Shouldn't dropping support for a version of PHP be considered a breaking change that results in a major point release
> 
> A BC break happens when your application breaks with a new release of a dependency.
> This isn't the case, so it can land in a minor release.
> 
> > It makes it difficult and incorrect version locking to be doing this in a Minor release.
> 
> Quite the contrary: this way you can upgrade your PHP version knowing nothing breaks

You are creating versions that prevent me from locking my dependencies to anything but patch releases, even if a Minor release still works. You are releasing Minor versions that break automated dependency management tools from working, like Renovate. This now requires me to manually track down compatible versions, which you don't even document in the release notes, and have to manually attempt to perform updates. At the very least your release notes should be clearly stating that PHP versions are being dropped.

Jean85:
> You are creating versions that prevent me from locking my dependencies to anything but patch releases, even if a Minor release still works.

No, the minor release will NOT work, because some dependencies needed here require PHP 8.4. PHPUnit 13.1 requires php-code-coverage v14, which is a new major that [requires PHP 8.4](https://github.com/sebastianbergmann/php-code-coverage/blob/24f1d7300e54e910197ee65e83d27a1e4bdc917e/composer.json#L32). Hence, Composer will refuse to install this minor, even if we didn't bump our requirement, making it pointless to leave as it is.

As you can see, PHPUnit is behaving the same way, bumping a minor while, under the hood, it's dropping support for a PHP version.

> You are releasing Minor versions that break automated dependency management tools from working, like Renovate.

No, it's not broken, it's working as intended. You CAN'T upgrade without jumping to 8.4, it's a requirement. The root blocker is two levels down, but it's still there.

> This now requires me to manually track down compatible versions, which you don't even document in the release notes, and have to manually attempt to perform updates.

No you don't. The tooling is behaving correctly, failing to install a dependency because you can't (yet).

timnolte:
You still didn't address the issue that you are not publishing in the release notes things like dropping PHP versions. This should be clear so people don't have to go digging through tons of releases and commits to try and find what version this happened on like I did in this case.

staabm:
this information is available in machine readable form in composer.json and you can easily obtain it using CLI commands.

all serious updating tools (dependabot, renovate,..) do this out of the box ..

PR #1089, opened 2026-03-28
Add `--shard-test-distribution` with a new `round-robin` method

rafal-althoff:
### Summary

- Add --shard-test-distribution option to control how tests are distributed across shards
- Implement round-robin distribution strategy as an alternative to the default sequential (contiguous blocks)
- Display selected distribution strategy in the output header
- Fully backward compatible — existing behavior is unchanged, sequential remains the default

### Motivation

The existing shard distribution assigns tests in contiguous blocks. This works well when test execution times are uniform, but becomes a bottleneck in projects that mix slow integration/feature tests with fast unit tests.

Most projects organize tests under separate namespaces (e.g. Tests\Feature, Tests\Unit), and PHPUnit loads them alphabetically. With sequential distribution, this means feature tests always land on the first shards and unit tests on the last — guaranteeing uneven execution times. This also means a single heavy feature test class can
end up on one specific shard, becoming the bottleneck for the entire pipeline.

Example: Consider a project with 20,000 tests split across 20 shards on 20 AWS ECS nodes:
- 10,000 feature tests (~10 min per 1,000 tests)
- 10,000 unit tests (~1 min per 1,000 tests)

With sequential distribution, the first 10 shards each get 1,000 feature tests (~10 min), while the last 10 shards each get 1,000 unit tests (~1 min). The pipeline waits for the slowest shard, so total wall time is ~10 minutes — with half the nodes sitting idle after just 1 minute.

With round-robin distribution, each shard receives a mix of feature and unit tests (~500 of each). Every node runs for roughly ~5.5 minutes, cutting total pipeline time nearly in half while fully utilizing all nodes.

### Usage

**Default behavior (sequential, unchanged)**
`vendor/bin/paratest --shard=1/20`
                                                                                                                
**Round-robin for balanced execution times**
`vendor/bin/paratest --shard=1/20 --shard-test-distribution=round-robin`

### DIFF
--- a/src/Options.php
+++ b/src/Options.php
@@ -7,6 +7,7 @@ namespace ParaTest;
 use Fidry\CpuCoreCounter\CpuCoreCounter;
 use Fidry\CpuCoreCounter\NumberOfCpuCoreNotFound;
 use InvalidArgumentException;
+use ParaTest\WrapperRunner\ShardDistribution;
 use PHPUnit\TextUI\Configuration\Builder;
 use PHPUnit\TextUI\Configuration\Configuration;
 use RuntimeException;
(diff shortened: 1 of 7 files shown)


Slamdunk:
I really like this proposal, thank you.

You are right that sequential distribution is the most likely to be suboptimal in a sharding context, but the round-robin feels working for you because your tests are well organized.

It would be good to also add a random one based on the seed that can be specificed with the dedicated option: I'll appreciate if you can propose a PR for that too.

But for now, this is good :+1:

Slamdunk (approved):

PR #1091, opened 2026-03-30
feat: add `random` shard test distribution

rafal-althoff:
## Summary

This PR adds a `random` distribution strategy to --shard-test-distribution, as suggested in the previous PR.

### Changes

- New `ShardDistribution::Random` enum case
- New `--shard-test-distribution-seed` option to control the seed for `mt_srand`/`shuffle` ? defaults to `0` (deterministic)
- Seed is validated (must be integer, only allowed with `random` distribution)
- Verbose output prints the seed when using random distribution
- Warning when sharding without `--functional`, since without it test classes run in full even if only some methods were assigned to a shard (pre-existing issue affecting all distribution strategies)
- Replaced hardcoded string values with `ShardDistribution` enum references for the default option value and in existing tests
- Full test coverage for the new option, including determinism, edge cases, and the warning

### 1. Default seed of `0` ? deterministic by default

Currently the seed defaults to `0`, which means `--shard-test-distribution=random` without an explicit seed always produces the same order. This is intentional for reproducibility (CI pipelines need consistent shard assignments), but it also means the distribution isn't truly "random".

An alternative would be to **require** `--shard-test-distribution-seed` when using `random` and throw a validation error if it's missing, but the end result would be the same for the end user. Would you prefer that approach, or is the current implicit default of `0` acceptable? Or potentially use actual random int, but it would be unusable for CI in my opinion. 

### 2. Sharding without `--functional` ? now warns

The distribution logic in `shardTests()` assigns individual test **methods** across shards. However, without `--functional` mode, the suite loader collapses results back to unique **class filenames** (`array_keys($files)` at `SuiteLoader.php:158-161`). When the runner passes that file to PHPUnit, **all** methods in the class get executed ? not just the ones assigned to this shard.

This means a class appears in a shard if **any** of its methods were selected, but then the entire class runs. Tests can end up executing on multiple shards, undermining the distribution. This is a pre-existing issue affecting all distribution strategies, not introduced by this PR.

This PR adds a warning when sharding is used without `--functional` to inform users. If you'd prefer a stricter approach (e.g. requiring `--functional` or force enabling it when sharding), let me know and I can adjust, but it could potentially break someones CI setup. 

Additionally, without `--functional` the progress output becomes misleading ? for example, if a shard selects 100 test methods across classes that collectively contain 150 methods, the counter will show `100/100` and then continue to `101/100`, `102/100`, etc. up to `150/100`, exceeding 100%. This is another symptom of the same root cause and further reinforces the need for the warning.


src/WrapperRunner/SuiteLoader.php
```diff
@@ -236,11 +241,25 @@ private function shardTests(TestSuite $suite): void
...
+    /**
+     * @param list<Test> $tests
+     *
+     * @return list<Test>
+     */
+    private function randomShardTests(array $tests, int $shards, int $current): array
+    {
+        mt_srand($this->options->shardDistributionSeed);
+        shuffle($tests);
+
```
Slamdunk, line 261: I don't like the idea to mess the random seed that may also be used by PHPUnit engine itself.
Can we leverage the `Random\Randomizer::shuffleArray` API please?

src/Options.php
```diff
@@ -371,8 +386,15 @@ public static function setInputDefinition(InputDefinition $inputDefinition): voi
...
-                'Distribution strategy for sharding: sequential (default) or round-robin',
-                'sequential',
+                'Distribution strategy for sharding: sequential (default), round-robin or random',
+                ShardDistribution::Sequential->value,
+            ),
+            new InputOption(
+                'shard-test-distribution-seed',
+                null,
+                InputOption::VALUE_REQUIRED,
+                'Seed for random shard test distribution. Defaults to 0, producing a deterministic order. Use different values to vary test distribution across runs',
```
Slamdunk: Any value procudes a deterministic order, we cold rephrase this as:
```suggestion
                'Seed for random shard test distribution. Defaults to the fixed value 0 to ensure reproducibility across different runs. Use different values to vary test distribution across runs',
```

src/Options.php
```diff
@@ -229,11 +230,24 @@ public static function fromConsoleInput(InputInterface $input, string $cwd): sel
...
+                'Invalid shard-test-distribution value: %s. Valid values are: sequential, round-robin, random',
                 $shardDistributionValue,
             ));
         }
 
+        $shardDistributionSeedValue = $options['shard-test-distribution-seed'];
+        unset($options['shard-test-distribution-seed']);
+        assert(is_string($shardDistributionSeedValue));
+        if (preg_match('/^-?\d+$/', $shardDistributionSeedValue) !== 1) {
+            throw new InvalidArgumentException('Shard test distribution seed must be an integer: ' . $shardDistributionSeedValue);
```
Slamdunk: This check is inconsistent with the requirements that follow when using very large numbers.
We can simplify this with:
```suggestion
        if ($shardDistributionSeedValue !== (string) (int) $shardDistributionSeedValue) {
            throw new InvalidArgumentException(sprintf('Shard test distribution seed must be an integer between %s and %s, value %s provided: ', PHP_INT_MIN, PHP_INT_MAX, $shardDistributionSeedValue));
```

src/Options.php
```diff
@@ -229,11 +230,24 @@ public static function fromConsoleInput(InputInterface $input, string $cwd): sel
         $shardDistribution = ShardDistribution::tryFrom($shardDistributionValue);
         if ($shardDistribution === null) {
             throw new InvalidArgumentException(sprintf(
-                'Invalid shard-test-distribution value: %s. Valid values are: sequential, round-robin',
+                'Invalid shard-test-distribution value: %s. Valid values are: sequential, round-robin, random',
```
Slamdunk: I'd like to use the enum values instead of hardcoded strings here, please

src/Options.php
```diff
@@ -371,8 +386,15 @@ public static function setInputDefinition(InputDefinition $inputDefinition): voi
                 'shard-test-distribution',
                 null,
                 InputOption::VALUE_REQUIRED,
-                'Distribution strategy for sharding: sequential (default) or round-robin',
-                'sequential',
+                'Distribution strategy for sharding: sequential (default), round-robin or random',
```
Slamdunk: I'd like to use the enum values instead of hardcoded strings here too, plese

Slamdunk (changes requested):
Thank you very much, again.
Few tweaks and we can 🚢  this

src/WrapperRunner/SuiteLoader.php
rafal-althoff: Good catch, I changed to andom\Randomizer::shuffleArray with Random\Engine\Mt19937 engine.

src/Options.php
rafal-althoff: I've applied it in the commit.

src/Options.php
rafal-althoff: I've applied it in the commit.

src/Options.php
rafal-althoff: I've changed it to use enum values. If someone will add more cases it wouldn't be necessary to change the message here.

src/Options.php
rafal-althoff: I've changed it to use enum values. If someone will add more cases it wouldn't be necessary to change the message here - one small difference I removed the `(default)` text, but It is added automatically at the end based on default InputOption value --help, so it looks like:
`--shard-test-distribution=SHARD-TEST-DISTRIBUTION            Distribution strategy for sharding: sequential, round-robin, random [default: "sequential"]`

Slamdunk:
> The distribution logic in `shardTests()` assigns individual test **methods** across shards.

I need to look this up because it doesn't sound right, but the PR is good, thank you

Slamdunk (approved):

rafal-althoff:
> > The distribution logic in `shardTests()` assigns individual test **methods** across shards.
> 
> I need to look this up because it doesn't sound right, but the PR is good, thank you

I can take a look at it as well, but I need 1-2 days, as I will be busy today. I believe that it is connected to run all checks over array values(test methods) instead of array keys(Test classes) if no `--functional` param is passed, but would need to investigate it further, as it is just my guess for now, i didn't focus on the issue, but noticed it when readding shard logic to pipelines in my project and forget to include functional flag at the beginning.

But basically if you have 3 test classes with 3 methods on each and use i.e. round robin, 
Then in theory each of 3 shards would have 1 test from each of the class right now, but then it is collapsed to classsd names and we run 9 tests instead of 3, and output thinks we run 3 tests, but after these 3 tests we will see 6 more rows with other test. Shard would run all tests anyway as the filtered methods are not taken into account since the code later on use test classes(without functional flag), so all tests will be run on such example instead of the ones that needs to be actually run. 

The correct approach would be to run in case of round-robin without functional: as each of 3 shards would get different class, and on functional level it should take 1 method from each of the 3 classes per shard. 

It is not catch by test as there are 1 test per file in test cases for shard tests.

PR #1095, opened 2026-04-03
bug: fix shard logic when `--functional` is disabled

rafal-althoff:
## Summary

- Fixed shard distribution to correctly distinguish between functional and non-functional mode: without `--functional`, sharding now distributes over whole test classes (keeping all methods of a class together), while `--functional` distributes over individual test methods
- Added `extractClassSuites()` method to `SuiteLoader` that extracts class-level `TestSuite` objects for file-based sharding
- Removed the misleading warning about sharding without `--functional` ? non-functional sharding now works correctly by design
- Rewrote shard distribution tests to use multi_method_tests fixtures (Alpha through Golf) with known method counts, verifying both file distribution and `testCount` per shard
- Added dedicated test cases for functional mode sharding (sequential, round-robin, random) that verify distribution over individual test methods

### DIFF
--- a/src/WrapperRunner/SuiteLoader.php
+++ b/src/WrapperRunner/SuiteLoader.php
@@ -98,10 +98,6 @@ final readonly class SuiteLoader
         $testSuite = (new TestSuiteBuilder())->build($this->options->configuration);
 
         if ($this->options->hasShard()) {
-            if (! $this->options->functional) {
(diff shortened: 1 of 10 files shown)


rafal-althoff:
@Slamdunk I've added draft with potential fix to the --shard option without functional issue. Didn't check it thoroughly yet, but it should work. Will test further tomorrow and let you know if it works fine.

Slamdunk (approved):

PR #1097, opened 2026-04-14
Fix JUnit XML missing test cases when merging suites in `--functional` mode

billypoke:
TestSuite::mergeWith() discarded the return value of recursive mergeWith() calls for overlapping nested suites. Since TestSuite is a readonly immutable class, the merged result was silently lost.

This caused JUnit XML output to report fewer <testcase> elements than actually ran when --functional mode split methods from the same class across multiple workers — the summary test count was correct, but CI tools that count <testcase> elements (e.g. Bitbucket) reported a lower number.

image

### DIFF
--- a/src/JUnit/TestSuite.php
+++ b/src/JUnit/TestSuite.php
@@ -125,7 +125,7 @@ final readonly class TestSuite
                 continue;
             }
 
-            $suites[$otherSuiteName]->mergeWith($otherSuite);
+            $suites[$otherSuiteName] = $suites[$otherSuiteName]->mergeWith($otherSuite);
         }
 
         ksort($suites);
--- a/test/Unit/JUnitTest.php
+++ b/test/Unit/JUnitTest.php
@@ -83,4 +83,31 @@ final class JUnitTest extends TestCase
         self::assertCount(2, $testSuite->suites['ParaTest\Tests\fixtures\github\GH997\SuccessfulTests']->cases);
         self::assertEquals(2, $testSuite->suites['ParaTest\Tests\fixtures\github\GH997\SuccessfulTests']->tests);
     }
+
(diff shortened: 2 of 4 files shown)


Slamdunk (approved):

Slamdunk:
Thank you

billypoke:
Thanks for the quick merge @Slamdunk 🚀 

~~Might need a second pass with a deeper look. It's a lot closer now, I wonder if some of my tests are empty or otherwise not being counted. Dataproviders perhaps?~~

image

~~Meanwhile In the text output for each shard~~

~~Shard 1 `OK (1314 tests, 3208 assertions)`~~
~~Shard 2 `OK (1314 tests, 3219 assertions)`~~
~~Shard 3 `OK (1314 tests, 3261 assertions)`~~

Confirmed this is a BB parsing error, not JUnit or Paratest or PHPUnit. Claude helped identify the issue

>PHPUnit wraps @dataProvider-driven test methods in a nested <testsuite> element. If a test class has only data-provided methods OR only plain
   methods, the structure is flat and Bitbucket counts correctly. But 7 of your classes mix both, so their direct <testcase> children get
  dropped.
>
>Example structure from CreatePaymentControllerTest (tests="4"):
> ```
>   <testsuite name="…CreatePaymentControllerTest" tests="4">
>     <testsuite name="…::itValidatesRequiredFields" tests="2">   ← Bitbucket counts this (2)
>       <testcase …/><testcase …/>
>     </testsuite>
>     <testcase name="itCanCreateASecurityDepositTransferPayment"/>   ← Lost
>     <testcase name="securityDepositTransferAmountsMustBeGreaterThanZero"/>   ← Lost
>   </testsuite>
> ```

PR #1103, opened 2026-04-29
Forward colors option to PHPUnit, even when it matches the default option value.

### LINKED ISSUE #1101: `--colors=never` is not forwarded to PHPUnit
| Q | A
| --- | ---
| ParaTest version | 7.22.3
| PHPUnit version | 13.1.7
| PHP version | 8.5.5

#### Summary

The `--colors=never` flag is not forwarded to PHPUnit because it coincides with the default value of the option. As a result, PHPUnit falls back to the `colors="true"` setting defined in `phpunit.dist.xml`.

Sources:
 
https://github.com/paratestphp/paratest/blob/bb620139a8326af832e1afbda0e1827885266c2c/src/Options.php#L195-L196

https://github.com/sebastianbergmann/phpunit/blob/0db0ddd8dcd1c0c01ad0d8e6d6233652af7a3aea/src/TextUI/Configuration/Configuration.php#L21-L24

#### Current behavior

The output remains colored even when `--colors=never` is specified.

```
php .\vendor\bin\paratest --colors=never
```

#### Expected behavior

The `colors` option should always be forwarded to PHPUnit, even when it matches the default value.

mickverm:
Fixes https://github.com/paratestphp/paratest/issues/1101

### DIFF
--- a/src/Options.php
+++ b/src/Options.php
@@ -191,13 +191,6 @@ final readonly class Options
         $functional = $options['functional'];
         unset($options['functional']);
 
-        assert(array_key_exists('colors', $options));
-        if ($options['colors'] === Configuration::COLOR_DEFAULT) {
-            unset($options['colors']);
-        } elseif ($options['colors'] === null) {
-            $options['colors'] = Configuration::COLOR_AUTO;
-        }
-
         assert(array_key_exists('coverage-text', $options));
         if ($options['coverage-text'] === null) {
             $options['coverage-text'] = 'php://stdout';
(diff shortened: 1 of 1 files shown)


Slamdunk (changes requested):
Thank you: could you please add a test, so we can be sure the bug doesn't get reintroduced in the future?

mickverm:
Yes, I was still working on it 😛

src/Options.php
```diff
@@ -191,13 +191,6 @@ public static function fromConsoleInput(InputInterface $input, string $cwd): sel
         $functional = $options['functional'];
         unset($options['functional']);
 
-        assert(array_key_exists('colors', $options));
-        if ($options['colors'] === Configuration::COLOR_DEFAULT) {
```
mickverm, line 195: I discovered that when I simply removed the 'if', the option fell back to the default nonetheless (when not passed as an argument) because of what was on line 700. Which meant it fell back to `--colors=never`, which would override the `colors="true"` configured in `phpunit.dist.xml`.

mickverm:
@Slamdunk could you point me in the direction of a related test? I'm having trouble figuring out how I could properly test this.

PR #1106, opened 2026-05-13
perf: avoid O(n²) array copies when merging worker TestResults

jorgsowa:
Replace 21 array_merge_recursive calls per result file with mutable accumulators and array_push, then construct TestResult once after the loop. Reduces complexity from O(N²×M) to O(N×M).

Doesn't look pretty, though. Maybe a later helper can be created to improve readability.


Slamdunk:
Please ensure the build pass on your local machine using `make` before publishing the changes.

jorgsowa:
Yes, sorry. That's why I marked PR as a draft.

PR #1122, opened 2026-07-28
Worker affinities for tests

pvandommelen:
This contains a new feature to allow assigning affinities to tests by implementing an interface.

In our test suite we are relying quite a bit on caching state between (integration) tests. This mostly includes fixtures and compiled DI containers. Currently we mostly do the bookkeeping for this state in memory. This obviously won't be very effective in paratest because workers don't share this state.

An obvious solution is to move this state to the file system. That is definitely the most important part of the solution, but has a few missing parts:
* The locking you would use would also cause workers to be idle
* Not all state/caches are easily serializable to disk.
* It requires disk IO, which makes it unsuitable for slow-but-not-that-slow caches.
* and probably most relevant, its actually quite complicated to do well.

This PR allows tests to be assigned an affinity. This can be used on top of, or as an alternative to, handling this using the file system. Tests with the same affinity are more likely to run within the same worker, while tests with a different affinity will likely run in different workers. Existing behaviour is unchanged. The tests will run as they do currently when this functionality is not used.

Three commits:
* First one splits up some logic in the `SuiteLoader` which allowed for some dead code to be cleaned up.
* Second creates `Suite` and `PendingTestQueue` classes to contain state so the complexity from the new functionality is kept out of the `WrapperRunner` class. The `SuiteLoader` also got a drive-by change to become an actual "loader" that creates something using a "load" method.
* Third actually implements the feature and adds tests and a section to the readme.

The feature adds a new `AffinityAware` interface that would be added to the public api of paratest. Users of paratest can optionally choose to implement this interface and its `getAffinity` method. `getAffinity` can return any string or `null`. Return `null` is equivalent to not implementing the interface.

For our test suite this achieves a speedup of somewhere between 10% and 45%. See table at the bottom. This should be taken with a grain of salt, as this is without any disk based state which I would expect to have a similar effect. Still not too bad for what amounts to reordering the tests.

Some implementation choices that are probably reasonable though they can be improved in the future:
* In the current implementation a test having no affinity is considered being part of a default affinity.
* The scheduler is kept simple. Affinity buckets with the most tests are prioritized. It balances workers by minimizing worker count per affinity and does not consider the number of remaining tests per affinity.

It is probably also important to note that there are a couple of features in paratest/phpunit that are affected, though only if this feature is used:
* Sharding and its various strategies. This affinity should take priority over sharding I believe, and sharding would only apply within an affinity.
* Test order. Similar, affinity takes priority and test order only affects the ordering within an affinity.
* Functional. Affinities are assigned per test class and can't be different per test method.
* ... maybe more?

In the end the change here is a bit more involved than probably should have been done without prior discussion. Let me know if you'd like a different approach or don't like this at all.

Minor: the `make` command wants to make an unrelated change to the composer.json file, this has been kept out of the PR.

LLM disclaimer: Claude was used for review, but not for planning, code, tests or this PR.

Performance benchmarks:

| Scenario                                               | --functional | --max-processes | time before | time with affinities |
| ------------------------------------------------------ | ----------- | ------ | ----------- | -------------------- |
| Component "graphql", 465 tests, 7 affinities                   | no          | 2      | 0m56 - 1m00   | 0m45 - 0m55            |
| Component "graphql", 465 tests, 7 affinities                   | yes         | 4      | 1m00 - 1m23   | 0m47 - 0m49            |
| Component "graphql", 465 tests, 7 affinities                   | no          | 4      | 0m45 - 0m59   | 0m41 - 0m48            |
| Component "customization", 240 tests, 23 affinities                  | no          | 4      | 3m59 - 4m56   | 3m14 - 3m33            |
| Component "customization", 240 tests, 23 affinities                  | no          | 12     | 4m56 - 5m01   | 2m46 - 2m52            |
| All incl simple unit tests, 3827 tests, 41 affinities | no          | 2      | 9m46        | 8m17                 |
| All incl simple unit tests, 3827 tests, 41 affinities | no          | 12     | 8m30        | 5m14                 |

### DIFF
--- a/README.md
+++ b/README.md
@@ -119,6 +119,52 @@ public function setUp(): void
 }
 ```
 
+## Test setup caching between tests
+
+Similar to the problem with single initialization in tests. Reusing (cached) data between tests using static
+variables will be less effective when tests are run on different worker processes.
+
+This has two solutions:
+* Share data using the file system. See a comparable solution using single initialization above.
+* Assign affinities to tests to hint that tests should run in the same worker. This is only a hint and not a guarantee.
+
+By default paratest will (without the `--functional` option) assign whole phpunit test cases to a worker.
+Assigning affinities will therefore only be useful if your state can be shared by multiple phpunit test cases.
+
+Assigning an affinity can be done by implementing the AffinityAware interface in your tests.
+
+```php
+final class ExpensiveSingleton
+{
+    private static self $instance;
+    
+    public static function get(): self
+    {
+        self::$instance ??= expensive_setup_logic();
+        return self::$instance;
+    }
+}
+
+//...
+
+use ParaTest\AffinityAware;
+
+final class MyTest extends TestCase implements AffinityAware
+{
+    private ExpensiveSingleton $singleton;
+    
+    protected function setUp(): void
+    {
+        $this->singleton = ExpensiveSingleton::get();
+    }
+
+    public function getAffinity(): string|null
+    {
+        return ExpensiveSingleton::class;
+    }
+}
+```
+
 ## Troubleshooting
 
 If you run into problems with `paratest`, try to get more information about the issue by enabling debug output via
--- a/src/AffinityAware.php
+++ b/src/AffinityAware.php
@@ -1,0 +1,16 @@
+<?php
+
+declare(strict_types=1);
+
+namespace ParaTest;
+
+/**
+ * Interface that can be implemented on phpunit test classes.
+ * Tests which share an affinity (any string) are more likely to be scheduled on the same worker process.
+ * This can improve performance when in-memory test setup or exclusive locks are used.
+ * This affinity is only a hint, and not a guarantee.
+ */
+interface AffinityAware
+{
+    public function getAffinity(): string|null;
+}
--- a/src/WrapperRunner/PendingTestQueue.php
+++ b/src/WrapperRunner/PendingTestQueue.php
@@ -1,0 +1,45 @@
+<?php
+
+declare(strict_types=1);
+
+namespace ParaTest\WrapperRunner;
+
+use function array_shift;
+use function count;
+
+/** @internal */
+final class PendingTestQueue
+{
+    /** @param list<non-empty-string> $pending */
+    public function __construct(
+        private array $pending,
+    ) {
+    }
+
+    /** @param list<non-empty-string> $tests */
+    public static function fromTests(array $tests): self
+    {
+        return new self($tests);
+    }
+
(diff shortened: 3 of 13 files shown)


Slamdunk:
This is a lot of code to add to the library and maintain for such a specific use case

> In our test suite we are relying quite a bit on caching state between (integration) tests.

This sounds very brittle. What I do in such (rare) scenarios is to persist and commit the cache:

* the test that is supposed to produce that cache checks every time it is equal to its output
* so the test that uses the cache is indipendent by eventual failings of the previous test and can run at full speed every time


For what I said, I'd rather decline this PR, I'm sorry you worked for it for nothing, next time please raise an issue before the coding session so it can be discussed thoroughly without wasting anyone time

pvandommelen:
Thanks for your quick response. It is not entirely unexpected to decline this as it adds a significant amount of surface area for paratest while its usage is situational.

It is a good reminder that committing these "cache" artifacts is an option. We should try to use that more. For this use case it is not a great fit as we currently have +-40 variations of a single cache item. Though it wouldn't be a bad plan to reduce that to +- 1.

It wouldn't cover the case for db fixtures though. The expensive state is data (schema + rows) materialized in a database. We currently track its state and truncate between tests.

I would expect this affinity to also improve some issues around locking of an external resource, though this is probably only minor and not something that we are currently running into.

Two potential followups:
* The first commit is independent of the feature and cleans up some dead code. If you want it you can cherry pick or I can open a separate PR.
* Would you be open to adding an extension point that would allow customizing the way in which tests are assigned to workers? Something like that would probably add more code and api surface area however. It could be more general and more powerful, which may allow it to support other use cases. If that sounds like a reasonable approach, I'll open an issue.

Slamdunk:
> * Would you be open to adding an extension point that would allow customizing the way in which tests are assigned to workers? Something like that would probably add more code and api surface area however. It could be more general and more powerful, which may allow it to support other use cases. If that sounds like a reasonable approach, I'll open an issue

That sounds reasonable indeed, but you are the only one asking for it (so far) and your use case is peculiar and non-standard, so right now I'm inclined to deny any further PR.

FYI, PHPUnit is about to add parallel exection support natively: https://github.com/sebastianbergmann/phpunit/pull/6784

pvandommelen:
Excellent point about native support in phpunit, I was not aware of that development. I can imagine that having a big impact on paratest.

For the extension point, I think something like that could also have been a solution for https://github.com/paratestphp/paratest/discussions/852. But fair, its use is likely only useful for some kind of project setups.

PR #1127, opened 2026-08-08
Fix: replace deprecated PHPUnit cache option

Guflly:
Fix https://github.com/paratestphp/paratest/issues/1126

PR #1128, opened 2026-08-12
fix: `--do-not-cache-result` CLI option is deprecated

### LINKED ISSUE #1126: "do-not-cache-result" CLI option is deprecated
| Q | A
| --- | ---
| ParaTest version | 7.24.0
| PHPUnit version |13.3.0
| PHP version|8.5.8

#### Summary

Similar to #1124, latest PHPUnit has more deprecations: a CLI option.

#### Current behavior

Paratest output:
```
There were 2 PHPUnit test runner deprecations:

1) The "--do-not-cache-result" CLI option is deprecated and will be removed in PHPUnit 14. Use "--do-not-record-test-run-history" instead.

2) The "--do-not-cache-result" CLI option is deprecated and will be removed in PHPUnit 14. Use "--do-not-record-test-run-history" instead.
```

#### Expected behavior

No deprecation message ;)

#### Additional information

I would hope that simply switching the option to the new one in `src/WrapperWorker.php` is enough.

j-schumann:
superseedes #1127 by also fixing the `TestBase` which did not isolate PHPUnit options set by individual tests from the options of the suite run.
This was previously the reason for the warning above `WrapperRunnerTest::testProcessIsolation` and caused `WrapperWorker.php` (running after  `WrapperRunnerTest.php`) to fail.

### DIFF
--- a/src/WrapperRunner/WrapperWorker.php
+++ b/src/WrapperRunner/WrapperWorker.php
@@ -134,7 +134,7 @@ final class WrapperWorker
             }
         }
 
-        $phpunitArguments[] = '--do-not-cache-result';
+        $phpunitArguments[] = '--do-not-record-test-run-history';
         $phpunitArguments[] = '--no-logging';
         $phpunitArguments[] = '--no-coverage';
         $phpunitArguments[] = '--no-output';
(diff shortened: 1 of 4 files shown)


j-schumann:
@Slamdunk, as far as I can see, those failures from the first CI run are flakiness caused by tests that assume output (like `EFIRS.W`) from parallel workers arriving in a fixed order, which may change through differing run times.

Slamdunk (approved):

